PR #7, opened 2022-10-06
Implement GstMeta to pass frame IDX through pipeline

tomskikh:


.gitignore
```diff
@@ -59,3 +59,7 @@ generated
 
 # local cached models
 models
+
+# gstsavantframemeta specific
+libs/gstsavantframemeta/_skbuild
+libs/gstsavantframemeta/dist
```
denisvmedyantsev, line 65: `dist` is ignored already

let's add `_skbuild` (just _skbuild) to `# Distribution / packaging` section

.dockerignore
```diff
@@ -7,3 +7,6 @@
 docker/
 .dockerignore
 Makefile
+libs/gstsavantframemeta/_skbuild
```
denisvmedyantsev, line 10: let's add these folders in general way, without `libs/gstsavantframemeta/`

denisvmedyantsev (changes requested):
Please, add links to gstsavantframemeta to the docs

denisvmedyantsev (approved):

PR #9, opened 2022-10-08
Fix batch size config

denisvmedyantsev:


savant/deepstream/nvinfer/element_config.py
```diff
@@ -123,6 +123,20 @@ def process(self, msg, kwargs):
...
+        if parse_result and (
+            model_config.batch_size,
+            model_config.gpu_id,
+            model_config.precision,
+        ) != (
+            parse_result['batch_size'],
+            parse_result['gpu_id'],
+            parse_result['precision'],
+        ):
+            model_config.engine_file = None
```
abramov-oleg, line 138: I think additional logging (at least on debug level) can be helpful here.

abramov-oleg (changes requested):

savant/deepstream/nvinfer/element_config.py
denisvmedyantsev: Added logger.info

abramov-oleg (approved):

PR #30, opened 2022-11-18
Fix memory leak on Jetson

denisvmedyantsev:


savant/deepstream/element_factory.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def create(self, element: PipelineElement) -> Gst.Element:
+        """Creates specified element.
+
+        :param element: Pipeline element to create.
+        :raises CreateElementException: Unknown element.
+        :return: Gst.Element.
+        """
+        # if isinstance(element, PyFuncElement):
+        #     return self.create_pyfunc_bin(element)
```
tomskikh, line 21: Please add a comment why we need `create_pyfunc_bin` and why we cannot use it now.

tomskikh:
With the new nvstrteammux we need to pass EOS event to the nvstreammuxsink pad before releasing the pad. Otherwise an error `App sequence error release pad called before push events`  occurs, and the last frame is lost. But there's an issue with passing EOS event to the sink pad - we cannot connect a new source since nvinfer throws an error `Source info not found for source 1. Maybe the GST_NVEVENT_PAD_ADDED event was never generated for the source.` (https://forums.developer.nvidia.com/t/the-new-nvstreammux-cannot-go-back-to-playing-state-after-eos/219046).

PR #45, opened 2023-01-11
Access to frames from PyFunc as OpenCV GpuMat

tomskikh:


libs/savantboost/savantboost/deepstream/nvsurfaceptr.cpp
```diff
@@ -102,10 +107,20 @@ void DSCudaMemory::UnMapCudaPtr() {
     cudaError_t err;
     Npp8u* _gpu_frame = nullptr;
     int status;
+
     if (_surface != nullptr)
         if (_surface->memType == NVBUF_MEM_SURFACE_ARRAY){
             if (_EglImageMapBuffer){
-                if (NvBufSurfaceUnMapEglImage(_surface, 0) !=0) GST_ERROR("Error NvBufSurfaceUnMapEglImage");
+                NvBufSurfaceParams &surface = _surface->surfaceList[_batch_id];
+                err = cudaGraphicsUnregisterResource(_pResource);
```
dorgun, line 115: If this problem was due to a missing synchronization and cudaGraphicsUnregisterResource solves it, it is worth removing the warning and branch 'else' with an alternative way in DSCudaMemory::GetMapCudaPtr. and DSCudaMemory::UnMapCudaPtr. And all the time use only mapping through eglImage.

savant/deepstream/opencv_utils.py
```diff
@@ -0,0 +1,132 @@
+"""Utils for working OpenCV representation of CUDA memory (cv2.cuda.GpuMat)."""
```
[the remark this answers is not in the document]
bwsw: @dorgun @tomskikh won't fix yet. It is ok to have it as is as long as it is correct and works as expected. We are not interested in more C++ code than minimally required to overcome the limitations of Python.

libs/savantboost/savantboost/deepstream/nvsurfaceptr.cpp
tomskikh: @dorgun This solves the issue with crashes on long videos. I removed mapping through CPU memory.

dorgun (approved):

PR #62, opened 2023-02-10
NvDsFrameMeta is extended and returns frame tags

dorgun:


savant/deepstream/meta/frame.py
```diff
@@ -21,6 +24,7 @@ def __init__(
         super().__init__()
         self.batch_meta = frame_meta.base_meta.batch_meta
         self.frame_meta = frame_meta
+        self.savant_frame_meta = nvds_frame_meta_get_nvds_savant_frame_meta(frame_meta)
```
denisvmedyantsev, line 27: The nvds_frame_meta_get_nvds_savant_frame_meta call will be executed for every frame, eg. in pyfunc. Are you sure it doesn't slow down the performance of the pipeline too much? Can we do it lazily?

savant/deepstream/meta/frame.py
```diff
@@ -48,6 +52,16 @@ def objects_number(self) -> int:
...
 
+    @property
+    def tags(self) -> Dict[str, Any]:
+        """Returns tags of frame. These tags are part of the meta information about
+        the frame that comes with the frames in the module.
+
+        :return: Dictionary with tags
+        """
+        source_meta = METADATA_STORAGE[self.source_id]
+        return source_meta.by_idx.get(self.savant_frame_meta.idx).tags
```
denisvmedyantsev, line 63: Did you try this with uridecodebin source? savant_frame_meta can be empty. Please, use the helper function metadata_get_frame_meta

denisvmedyantsev (changes requested):

savant/deepstream/meta/frame.py
dorgun: I moved the call of nvds_frame_meta_get_nvds_savant_frame_meta and now it will only be called once if we have not yet received GstFrameMeta

savant/deepstream/meta/frame.py
dorgun: I used metadata_get_frame_meta to get the GSTFrameMeta

savant/deepstream/meta/frame.py
```diff
@@ -48,6 +52,24 @@ def objects_number(self) -> int:
...
         return self.frame_meta.num_obj_meta
 
+    @property
+    def tags(self) -> Dict[str, Any]:
+        """Returns tags of frame. These tags are part of the meta information about
+        the frame that comes with the frames in the module.
+
+        :return: Dictionary with tags
+        """
+        if self.gst_frame_meta is None:
```
denisvmedyantsev, line 62: We can single out the lazy load of gst_frame_meta and use it in the future for other source frame properties in the future. Up to you.

denisvmedyantsev (approved):

savant/gstreamer/metadata.py
```diff
@@ -1,58 +1,67 @@
...
 from dataclasses import dataclass, field
+from types import MappingProxyType
 from typing import Any, Dict, Optional, Union
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_FRAMERATE = '30/1'
 
 
+class OnlyExtendedDict(UserDict):
```
abramov-oleg, line 13: It's possible for the user to change values in this type of dict by removing keys first via `__delitem__()`, `pop()`, `popitem()`, `clear()`.

dorgun: I have overridden these methods

abramov-oleg (approved):

PR #64, opened 2023-02-15
53 move to ds 62

denisvmedyantsev:


savant/gstreamer/utils.py
```diff
@@ -1,7 +1,9 @@
 """GStreamer utils."""
 import logging
+import os
 from contextlib import contextmanager
-from typing import Any, Callable, Dict
+from datetime import datetime
+from typing import Any, Callable, Dict, Optional
```
tomskikh, line 6: These are unused import statements.

tomskikh (approved):

savant/gstreamer/utils.py
denisvmedyantsev: fixed

PR #68, opened 2023-02-21
49 implement draw element artist using gpumat opencv

abramov-oleg:


Makefile
```diff
@@ -28,6 +28,7 @@ build-samples:
 	DOCKER_BUILDKIT=1 docker build \
 	--target samples \
 	--build-arg DEEPSTREAM_VERSION=$(DEEPSTREAM_VERSION) \
+	--build-arg PYDS_TAG=v$(PYDS_VERSION)-unmap \
```
denisvmedyantsev, line 31: There is no PYDS_VERSION here, the default arg value in the dockerfile is enough. Please, remove it.

scripts/common.py
```diff
@@ -13,8 +13,8 @@
 from savant.utils.platform import is_aarch64, get_l4t_version  # noqa: F401
 
 # docker registry to use with scripts, set to "None" to use local images
-DOCKER_REGISTRY = 'ghcr.io/insight-platform'
-# DOCKER_REGISTRY = None
+# DOCKER_REGISTRY = 'ghcr.io/insight-platform'
```
denisvmedyantsev, line 16: The current ghcr registry is outdated (until the next update), but keeping the original config is best.

denisvmedyantsev (changes requested):

scripts/common.py
abramov-oleg: Fixed.

Makefile
abramov-oleg: Fixed.

savant/utils/artist/artist_gpumat.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        frame_left, frame_top = origin
+        if frame_left >= self.width or frame_top >= self.height:
+            return
+
+        img_h, img_w = img.shape[:2]
+        if frame_left + img_w < 0 or frame_top + img_h < 0:
+            return
+
+        self.__init_overlay()
```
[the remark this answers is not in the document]
abramov-oleg: The intermediate overlay array is used to limit the number of `cv2.cuda.alphaComp()` calls by accumulating the results of drawing text boxes, polygons and adding overlay elements.

Presumably, this leads to less waiting for host-device transfers and increased performance.

tomskikh (approved):

denisvmedyantsev (approved):

PR #69, opened 2023-02-21
Frame RoI change

denisvmedyantsev:


savant/deepstream/buffer_processor.py
```diff
@@ -105,13 +106,33 @@ def prepare_input(self, buffer: Gst.Buffer):
...
+                # skip primary object for now, will be added later
+                if obj_key == PRIMARY_OBJECT_LABEL:
+                    bbox = (
+                        obj_meta['bbox']['xc'],
+                        obj_meta['bbox']['yc'],
+                        obj_meta['bbox']['width'],
+                        obj_meta['bbox']['height'],
+                    )
+                    # if not a full frame then correct primary object
+                    if bbox != (0.5, 0.5, 1, 1):
```
dorgun, line 131: I suggest adding a warning log message because if the first object is not equal to the whole frame, it is abnormal.

savant/deepstream/meta/frame.py
```diff
@@ -40,6 +42,22 @@ def objects(self) -> Iterator[ObjectMeta]:
...
+        if not self._primary_obj:
+            for obj_meta in self.objects:
+                if obj_meta.is_primary:
+                    self._primary_obj = obj_meta
+                    break
+        return self._primary_obj.bbox
+
+    @roi.setter
+    def roi(self, value: BBox):
+        self._primary_obj.bbox.x_center = value.x_center
```
dorgun, line 56: _primary_obj can be None. 
```suggestion
        self.roi.x_center = value.x_center
```

savant/meta/bbox.py
```diff
@@ -48,10 +50,48 @@ def left(self, value: float):
         left corner."""
         self.x_center = value + 0.5 * self.width
 
+    def scale(self, scale_x: float, scale_y: float):
+        """Scales BBox.
+
+        :param scale_x:
```
dorgun, line 56: Please add a description of the parameters. Something like "The scaling factor to apply along the x-axis"

dorgun (changes requested):

savant/deepstream/buffer_processor.py
denisvmedyantsev: Why is it abnormal? For example, I want to remove a bottom area from processing (there is the camera name and date), and I want to do it for cascaded pipelines. Why not?

savant/deepstream/meta/frame.py
denisvmedyantsev: It can be None by definition, but our current pipeline logic relies on the existence of the primary object. So if we don't have the primary object we will get an error and it's OK.

savant/meta/bbox.py
denisvmedyantsev: fixed

savant/deepstream/meta/frame.py
dorgun: https://github.com/insight-platform/Savant/blob/41-roi-change/savant/deepstream/pyfunc.py#L27

If you try to use `frame_meta.roi = BBox(100,100,200,200)` in your pyfunc, you will encounter an error indicating that self._primary_obj is None. However, in the deepstream meta, it will be defined.

dorgun (changes requested):

savant/deepstream/buffer_processor.py
denisvmedyantsev: You are right. Now the frame and the primary RoI object are the same and it's a bit confusing. With this code, I'm just providing a way to change the RoI. I think we need to decide how to deal with it, but not in this task. I have some thoughts, which we'll discuss in a meeting.

savant/deepstream/meta/frame.py
denisvmedyantsev: fixed (the case is added to the test)

dorgun (approved):

PR #70, opened 2023-02-22
66 quality and bitrate configuration for savant output frame

abramov-oleg:


docs/source/main_concepts/module.rst
```diff
@@ -50,6 +50,46 @@ in dynamic parameter storage.
...
+
+1. ``parameters.output_frame.codec`` config node determines which codec will be used to encode output frames. Possible values are **h264**, **h265**, **jpeg**, **png** and **raw-rgba** to skip encoding. For example:
+
+  .. code-block:: YAML
+
+    parameters:
+      output_frame:
+        codec: h264
+
+2. ``parameters.output_frame.encoder_params`` config node allows passing additional encoder-specific parameters to the encoder pipeline element. These are:
```
dorgun, line 66: The documentation should describe the parameters for each codec with which the user can set the encoding quality, so that the user does not have to run gst-inspect-1.0 to find them.

dorgun (changes requested):

docs/source/main_concepts/module.rst
abramov-oleg: Listed encoder elements properties explicitly.

dorgun (approved):

Issue #171, opened 2023-04-16
Build 0.2.1 release


bwsw:
As stated, we must build it and make it available on GitHub Registry.

Ready to build after the resolution of #161 and the merge of  #162  #167 #170 MRs. 

- [x] Remove samples docker image

Issue #161, opened 2023-04-13
Deploy Savant python code as wheel package


bwsw:
The Savant code must be deployed to PyPi to support installation as a dependency for development purposes. 
Within the docker image, Savant also must be used as a wheel to support IDEs which can do a docker container inspection to discover dependencies.

Auxiliary libraries must also be uploaded on PyPi as wheel packages to support IDEs.

Issue #146, opened 2023-04-08
Document how to setup the development environment


bwsw:
We need a tutorial and dev environment on how to develop for Savant. It looks like the simplest way to develop for Savant is to use "Remote Environment", which can be found in IntelliJ Professional tools. 
They allow execution of the source code within the docker container created from a certain image.

1. The user downloads the container;
2. She uses the container as a remote environment;
3. The PyCharm introspects it and extracts all the libraries and they become available in the IDE (Savant must be installed as EGG);
4. The user creates YAML pipeline in a project root;
5. The user creates pyfuncs;
6. The user specifies "main.py" where she specifies YAML to load, iterator to get [(frame, meta), ...] and class to receive results.
7. On "run", the pipeline is launched and works until the iterator yields the data.
8. The logs are displayed in the Run console.
9. The HowTo explains how to pack the project to Savant.

It enables the creation of test suites and interactive development.

Pseudocode:

```python
pipeline = Pipeline("config.yml")
       .process(source="name", fps=1/30, src_iterator1, sink_cls_instance2)
       .process(source="other-name", fps=-1, src_iterator2, sink_cls_instance2)
       .run(parallel=True) # blocking
```

* Pipeline,
* Element (image[, extra_meta: MetaObject]) - in, out
* EOS
* OpenCVMat | JPEGBuffer
* MetaObject

I tried to use Remote Docker Runtime in PyCharm Professional, it works for local Docker daemon well, but doesn't work well for remote SSH-accessed Docker.

![image](https://user-images.githubusercontent.com/15047882/230781070-48a828d2-83dc-4260-8761-8f54e65b6307.png)

![image](https://user-images.githubusercontent.com/15047882/230781105-f02dc2f4-5cca-4a55-9d8c-128756946cf7.png)

![image](https://user-images.githubusercontent.com/15047882/230781157-0bf168ff-2d7f-4fbf-b1f9-279f00d0e72a.png)

![image](https://user-images.githubusercontent.com/15047882/230781202-3e8c31fb-a385-4281-aa70-d4e5ead52ce2.png)

How to craft #147

bwsw:
Implement as part of #160

denisvmedyantsev:
To run savant with docker interpreter in pycharm do not forget to add `--gpus=all` in docker run options.
![Screenshot from 2023-04-21 12-10-22](https://user-images.githubusercontent.com/44088010/233673340-7665c217-1c69-4477-9051-e498a3d0b5f7.png)

PR #187, opened 2023-04-21
Deploy savant as package

denisvmedyantsev:
#146, #161, #171


samples/template/README.md
```diff
@@ -0,0 +1,20 @@
+# Savant module template
+
+Start your own project with this template.
+
+## Develop with IDE
+
+Copy the template, rename. Open template with your favorite IDE.
```
dorgun, line 7: If you copy and rename template, when you try to run.py an error occurs because it cannot find overlay and custom_pyfunc. In module.yml the path is samples.template which does not exist.

dorgun (changes requested):

samples/template/README.md
denisvmedyantsev: fixed

samples/template/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+### PyCharm Professional
+1. [Configure a docker interpreter](https://www.jetbrains.com/help/pycharm/using-docker-as-a-remote-interpreter.html) using savant-deepstream container `ghcr.io/insight-platform/savant-deepstream:latest`.
```
dorgun, line 10: We should use the Docker image of the module because if the user uses third-party packages specified in the requirements.txt file, they will not be present in the base image.
Therefore, when setting up PyCharm, the user should select the appropriate Docker file and 'context folder', which is the project folder, and choose 'build' instead of 'pull' for the image.
Unfortunately, PyCharm does not automatically collect information about newly installed packages in the Docker container. However, there is an option to manually scan for new packages. To do this, go to the settings and search for "rescan", then navigate to Plugins -> Python -> Rescan Available Python Modules and Packages and set the hotkey. After adding a new package to requirements.txt, simply press the specified hotkey to rebuild the Docker image and update the packages.

dorgun (changes requested):

dorgun (approved):

PR #211, opened 2023-05-19
209: Add a check to recognize when encoding device is not present or limited (GeForce)

### LINKED ISSUE #209: Add a check to recognize when encoding device is not present
Not all Nvidia hardware support encoding. If the pipeline uses H264, HEVC, or other encoding technology that uses NVENC, we need to ensure that NVENC is present. In case it is not present like our customer encountered:

image

The system must react properly, terminating and suggesting to change output frame properties accordingly to `raw-rgba` or other in-software technology without NVENC. 

Respectively, when the system cannot allocate additional hardware ENCODER (like with GeForce), it must react properly, resulting in an understandable message and terminate clearly.

tomskikh:
Closes #209

### DIFF
--- a/gst_plugins/python/avro_video_demux.py
+++ b/gst_plugins/python/avro_video_demux.py
@@ -140,6 +140,7 @@ class AvroVideoDemux(LoggerMixin, Gst.Element):
         self.source_eviction_interval = DEFAULT_SOURCE_EVICTION_INTERVAL
         self.last_eviction = 0
         self.source_lock = Lock()
+        self.is_running = False
         self.expiration_thread = Thread(target=self.eviction_job, daemon=True)
         self.store_metadata = False
         self.max_parallel_streams: int = 0
@@ -166,6 +167,7 @@ class AvroVideoDemux(LoggerMixin, Gst.Element):
             and new != Gst.State.NULL
             and not self.expiration_thread.is_alive()
         ):
+            self.is_running = True
             self.expiration_thread.start()
 
     def do_get_property(self, prop):
@@ -213,6 +215,12 @@ class AvroVideoDemux(LoggerMixin, Gst.Element):
             buffer.get_size(),
             buffer.pts,
         )
+        if not self.is_running:
+            self.logger.info(
+                'Demuxer is not running. Skipping buffer with timestamp %s.',
+                buffer.pts,
+            )
+            return Gst.FlowReturn.OK
         frame_meta_mapinfo: Gst.MapInfo
         result, frame_meta_mapinfo = buffer.map_range(0, 1, Gst.MapFlags.READ)
         assert result, 'Cannot read buffer.'
@@ -229,6 +237,7 @@ class AvroVideoDemux(LoggerMixin, Gst.Element):
             result = self.handle_eos(message)
         else:
             self.logger.error('Unknown schema "%s"', schema_name)
+            self.is_running = False
             result = Gst.FlowReturn.ERROR
 
         buffer.unmap(frame_meta_mapinfo)
@@ -265,6 +274,7 @@ class AvroVideoDemux(LoggerMixin, Gst.Element):
                     self.max_parallel_streams
                     and len(self.sources) >= self.max_parallel_streams
                 ):
+                    self.is_running = False
                     error = (
                         f'Failed to add source {source_id!r}: reached maximum '
                         f'number of streams: {self.max_parallel_streams}.'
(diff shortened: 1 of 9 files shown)


gst_plugins/python/avro_video_demux.py
```diff
@@ -263,14 +265,21 @@ def handle_video_frame(self, message: Dict, buffer: Gst.Buffer) -> Gst.FlowRetur
...
+                    frame = inspect.currentframe()
+                    propagate_gst_error(
+                        gst_element=self,
+                        frame=frame,
+                        file_path=__file__,
+                        domain=Gst.StreamError.quark(),
+                        code=Gst.StreamError.DEMUX,
+                        text=error,
+                    )
+                    return Gst.FlowReturn.ERROR
```
tomskikh, line 282: Sometimes the module hangs when started simultaneously with `max_parallel_streams+1` sources.

```
2023-05-19 11:39:34,005 [ERROR] [savant.avro_video_demux] [Dummy-4] Failed to add source 'city-traffic-015ec1b103e3': reached maximum number of streams: 2.
2023-05-19 11:39:34,006 [ERROR] [savant.gstreamer.runner] [Thread-3] Received error "gst-stream-error-quark: Failed to add source 'city-traffic-015ec1b103e3': reached maximum number of streams: 2. (9)" from avro_video_demux+avrovideodemux0.
[NvMultiObjectTracker] De-initialized
```

tomskikh: The module hangs when it tries to add elements or request pads after `shutdown()` was called. Need to check pipeline is running before adding elements or requesting pads.

savant/deepstream/encoding.py
```diff
@@ -0,0 +1,78 @@
...
+    pipeline: Gst.Pipeline = Gst.Pipeline.new()
+
+    converter_props = {}
+    if not is_aarch64():
+        converter_props['nvbuf-memory-type'] = int(pyds.NVBUF_MEM_CUDA_UNIFIED)
+
+    elements = [
+        PipelineElement(
+            'videotestsrc',
+            properties={'num-buffers': 50},
```
abramov-oleg: Do we need 50 buffers here? Maybe 1 buffer is enough?

savant/deepstream/encoding.py
```diff
@@ -0,0 +1,78 @@
...
+        pipeline.add(gst_element)
+        if last_gst_element is not None:
+            logger.debug('Linking %r -> %r', last_gst_element.name, gst_element.name)
+            assert last_gst_element.link(gst_element)
+        last_gst_element = gst_element
+
+    with GstPipelineRunner(pipeline) as runner:
+        while runner._is_running:
+            time.sleep(0.1)
+        if runner._error is not None:
```
abramov-oleg, line 70: Is it possible to check for a specific error message here?

savant/deepstream/encoding.py
```diff
@@ -0,0 +1,78 @@
...
+        else:
+            gst_element = GstElementFactory.create_element(element)
+        logger.debug('Created element %r', gst_element.name)
+        pipeline.add(gst_element)
+        if last_gst_element is not None:
+            logger.debug('Linking %r -> %r', last_gst_element.name, gst_element.name)
+            assert last_gst_element.link(gst_element)
+        last_gst_element = gst_element
+
+    with GstPipelineRunner(pipeline) as runner:
```
abramov-oleg: There's a new class `NvDsPipelineRunner`, can we use it here and rely on the specific error message that it builds for the case of lacking encoder sessions?

savant/deepstream/pipeline.py
```diff
@@ -376,9 +405,11 @@ def _add_input_converter(
             source_info.before_muxer.append(video_converter)
 
         source_info.before_muxer.append(nv_video_converter)
+        self._check_pipeline_is_running()
         # TODO: send EOS to video_converter on unlink if source didn't
         assert new_pad.link(video_converter_sink) == Gst.PadLinkReturn.OK
 
+        self._check_pipeline_is_running()
```
abramov-oleg, line 412: If I understand correctly, the method `_check_pipeline_is_running()` seems to be called repeatedly to account for possible changes to the `self._is_running` flag from parallel threads.

Isn't it better to use some synchronization mechanism to ensure immutability for this flag during the `_add_input_converter()` method (and other same casese) call?

savant/deepstream/encoding.py
tomskikh: Yes, 1 buffer would be enough.

savant/deepstream/encoding.py
tomskikh: We don't need it. The error message would be logged anyway.

savant/deepstream/pipeline.py
tomskikh: We can add a synchronization on a case when user stops the module. But when the module fails with an error synchronization would be difficult to add.

abramov-oleg (approved):

PR #227, opened 2023-05-31
120: Fix frame loss on dealer router sockets

### LINKED ISSUE #120: Frame loss with a specific video file in "Source > Module" bundle
There is frame loss with the current Dealer-Router configuration for Source-Module. 

There is no loss if set 1) sync=1 for the source (decrease FPS significantly) or 2) modify HWM (requires investigation, see https://github.com/zeromq/jeromq/issues/703) or 3) change ZMQ schema to REQ-REP.

To reproduce run `./scripts/run_source.py videos --source-id=0 ../vlf-pipelines/data/bottle_defect_detector/test.mp4` and module with minimal config, and compare log messages `Processed N frame..` for source and module (2494 vs 2478 in my case, 16 frames loss).

module config
```
name: test
parameters:
  batch_size: 1
```

tomskikh:
Closes #120

### DIFF
--- a/adapters/gst/gst_plugins/python/zeromq_sink.py
+++ b/adapters/gst/gst_plugins/python/zeromq_sink.py
@@ -9,7 +9,7 @@ from gst_plugins.python.zeromq_properties import (
     ZEROMQ_PROPERTIES,
 )
 from savant.gstreamer import GObject, Gst, GstBase
-from savant.gstreamer.utils import propagate_gst_setting_error
+from savant.gstreamer.utils import propagate_gst_error, propagate_gst_setting_error
 from savant.utils.logging import LoggerMixin
 from savant.utils.zeromq import (
     Defaults,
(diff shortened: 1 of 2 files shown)


adapters/gst/gst_plugins/python/zeromq_sink.py
```diff
@@ -177,6 +178,16 @@ def do_render(self, buffer: Gst.Buffer):
...
         """Stop source."""
+
+        if self.socket_type == SenderSocketTypes.DEALER:
+            self.logger.info('Sending End-of-Stream message to socket %s', self.socket)
+            self.sender.send_multipart([END_OF_STREAM_MESSAGE])
+            self.logger.info(
+                'Waiting for End-of-Stream message confirmation from socket %s',
+                self.socket,
+            )
+            self.sender.recv()
```
bwsw: @tomskikh I would propose setting a reasonable timeout here (like 10 seconds). And to report with ok (0) zero code or err (1) and message if timeout happened.

abramov-oleg (approved):

adapters/gst/gst_plugins/python/zeromq_sink.py
tomskikh: @bwsw Added a timeout.

PR #237, opened 2023-06-13
Implement conditional encoding/drawing based on per-stream tags

### LINKED ISSUE #84: Implement conditional encoding/drawing based on per-stream tags
If the frame doesn't have a specified tag, the frame doesn't pass to the encoder. Similarly, if the frame doesn't have a specified tag (other than the encoding tag), the draw element skips it.

The feature can lead to generating of 'sparse' video streams which are partially filled with frames (e.g. only key-frames are configured with tags).

Why it is required?

We may need to debug and visualize only certain streams; for the others, we don't want to encode and draw to save resources. To implement that, we may use a custom source/transit adapter, which injects additional tags into streams that must be encoded and visualized. 

Similarly, only when something useful occurs in the stream a `pyfunc` may be utilized, which will inject a required tag causing the frames to be drawn and encoded to the output. If there is no valuable information within the frame, it is not drawn or encoded.

tomskikh:


savant/gstreamer/utils.py
```diff
@@ -115,3 +116,25 @@ def gst_buffer_from_list(data: List[bytes]) -> Gst.Buffer:
     for item in data:
         buffer = buffer.append(Gst.Buffer.new_wrapped(item))
     return buffer
+
+
+class RequiredPropertyError(Exception):
+    """Raised when required property is not set."""
+
+    pass
```
abramov-oleg: `pass` statement is not necessary, a docstring is sufficient to finish class definition.

savant/deepstream/pipeline.py
```diff
@@ -787,10 +738,21 @@ def _draw_on_frame_probe(
         buffer: Gst.Buffer = info.get_buffer()
         nvds_batch_meta = pyds.gst_buffer_get_nvds_batch_meta(hash(buffer))
         for nvds_frame_meta in nvds_frame_meta_iterator(nvds_batch_meta):
-            self._draw_func(nvds_frame_meta, buffer)
+            if self._can_draw_on_frame(nvds_frame_meta):
```
abramov-oleg, line 770: Propose to add debug log to record why the drawfunc call was skipped.

savant/config/schema.py
```diff
@@ -236,6 +236,13 @@ class DrawFunc(PyFunc):
     kwargs: Optional[Dict[str, Any]] = None
     """Class initialization keyword arguments."""
 
+    when_tagged: Optional[str] = None
```
abramov-oleg: The name `when_tagged` seems a bit confusing, in my opinion.

This property is the tag, so I think something like `enabler_tag` should be more clear.

gst_plugins/python/frame_tag_filter.py
```diff
@@ -0,0 +1,233 @@
...
+            self.logger.debug(
+                'savant_frame_meta=%s, frame_idx=%s, frame_pts=%s',
+                savant_frame_meta,
+                frame_idx,
+                frame_pts,
+            )
+            frame_meta = get_source_frame_meta(self.source_id, frame_idx, frame_pts)
+            if self.tag in frame_meta.tags:
+                self.logger.debug('Frame PTS=%s has tag "%s"', frame_pts, self.tag)
+                return None
```
abramov-oleg, line 194: Why return and not continue?

gst_plugins/python/frame_tag_filter.py
```diff
@@ -0,0 +1,233 @@
...
+        tagged = not_tagged_buffers is None
+        if tagged != self.last_tagged:
+            self.logger.debug('Switching to tagged=%s', tagged)
+            event = build_stream_part_event(next(self.part_id_gen), tagged)
+            self.src_pad_not_tagged.push_event(event)
+            self.src_pad_tagged.push_event(event)
+
+        self.last_tagged = tagged
+        if tagged:
+            return self.push_buffer(self.src_pad_tagged, buffer)
```
abramov-oleg, line 163: Why is there no error logging similar to the one below for the non_tagged_buffers?

savant/config/schema.py
bwsw: @tomskikh @abramov-oleg 
Let it be:
```
condition:
  tag: tagname
```

savant/deepstream/pipeline.py
tomskikh: Added logging into method `_can_draw_on_frame`.

gst_plugins/python/frame_tag_filter.py
tomskikh: Added logging into method `push_buffer`.

gst_plugins/python/frame_tag_filter.py
```diff
@@ -0,0 +1,233 @@
...
+                    'Failed to push buffer PTS=%s: %s',
+                    not_tagged_buffer.pts,
+                    ret,
+                )
+                return ret
+
+        return Gst.FlowReturn.OK
+
+    def parse_buffer(self, buffer: Gst.Buffer) -> Optional[List[Gst.Buffer]]:
+        """Parse buffer and return list of not tagged buffers or None if buffer is tagged."""
```
abramov-oleg, line 173: Please add a comment, that the DS batch is expected to have only 1 frame at this moment, and `non_tagged_buffers` list is just a precaution

savant/deepstream/source_output.py
```diff
@@ -239,3 +369,42 @@ def dest_resolution(self, source_info: SourceInfo) -> Resolution:
...
+
+    if not output_frame:
+        return SourceOutputOnlyMeta()
+
+    codec = CODEC_BY_NAME[output_frame['codec']]
+    if codec == Codec.RAW_RGBA:
+        return SourceOutputRawRgba(frame_params=frame_params)
+
+    encoder_params = output_frame.get('encoder_params', {})
+    condition = FrameProcessingCondition(**(output_frame.get('condition') or {}))
```
abramov-oleg, line 388:
```suggestion
    condition = FrameProcessingCondition(**output_frame.get('condition', {}))
```

abramov-oleg (changes requested):
Also need to update tests.

savant/deepstream/source_output.py
tomskikh: I've added `or` in case `output_frame` contains `condition=None`.

abramov-oleg (approved):

PR #250, opened 2023-06-16
Replace BBox structs with rust implementation

### LINKED ISSUE #207: Change python bboxes with savant-rs bboxes
Implemented BBox, RBBox.

Get rid of (implemented in the bbox/rbbox):
* https://github.com/insight-platform/Savant/blob/develop/savant/utils/artist/artist_gpumat.py#L360
* https://github.com/insight-platform/Savant/blob/develop/savant/meta/bbox.py#L88
* https://github.com/insight-platform/Savant/blob/develop/savant/converter/scale.py

Objective:
* measure how the performance changes.

abramov-oleg:
closes #207


savant/deepstream/drawfunc.py
```diff
@@ -113,20 +113,20 @@ def _draw_bounding_box(
         else:
             padding = (0, 0, 0, 0)
         artist.add_bbox(
-            bbox=obj_meta.bbox,
-            border_color=spec.border_color.rgba,
-            border_width=spec.thickness,
-            bg_color=spec.background_color.rgba,
-            padding=padding,
+            obj_meta.bbox,
```
dorgun, line 116: Why are changes like this made? Passing by key is more visual, allows easier refactoring, and allows you to sometimes find bugs with the wrong call.

savant/deepstream/pipeline.py
```diff
@@ -590,6 +594,14 @@ def update_frame_meta(self, pad: Gst.Pad, info: Gst.PadProbeInfo):
             )
             frame_idx = savant_frame_meta.idx if savant_frame_meta else None
             frame_pts = nvds_frame_meta.buf_pts
+
+            self._logger.debug(
+                'Preparing output for frame of source %s with IDX %s and PTS %s.',
+                source_id,
+                frame_idx,
+                frame_pts,
+            )
```
dorgun:
```suggestion
            if self._logger.isEnabledFor(logging.DEBUG):
                self._logger.debug(
                    'Preparing output for frame of source %s with IDX %s and PTS %s.',
                    source_id,
                    frame_idx,
                    frame_pts,
                )
```

savant/deepstream/pipeline.py
```diff
@@ -613,24 +625,42 @@ def update_frame_meta(self, pad: Gst.Pad, info: Gst.PadProbeInfo):
...
+                    )
+                    if not bbox.almost_eq(dest_res_bbox, 1e-6):
+                        if self._logger.isEnabledFor(logging.DEBUG):
+                            self._logger.debug(
+                                'Adding primary object, bbox %s != dest res bbox %s.',
+                                bbox,
+                                dest_res_bbox,
+                            )
+                    elif obj_meta['attributes']:
+                        if self._logger.isEnabledFor(logging.DEBUG):
```
dorgun: I think there is an unnecessary check, there is a simple line in the log

savant/deepstream/pipeline.py
```diff
@@ -613,24 +625,42 @@ def update_frame_meta(self, pad: Gst.Pad, info: Gst.PadProbeInfo):
...
+                            self._logger.debug(
+                                'Adding primary object, attributes not empty.'
+                            )
+                    else:
+                        if self._logger.isEnabledFor(logging.DEBUG):
```
dorgun: I think there is an unnecessary check, there is a simple line in the log

savant/deepstream/utils/object.py
```diff
@@ -154,32 +139,147 @@ def nvds_generate_obj_uid(
...
+    """Update bbox values for specified object meta
+    into Deepstream meta structures."""
+    if isinstance(bbox, BBox):
+        set_aligned_bbox_for_obj_meta(obj_meta, bbox)
+
+    elif isinstance(bbox, RBBox):
+        update_rotated_bbox_for_obj_meta(obj_meta, bbox)
+
+
+def update_rotated_bbox_for_obj_meta(
```
dorgun: Unnecessary function, you can add code to nvds_upd_obj_bbox. In any case only nvds_upd_obj_bbox should be used.

savant/deepstream/utils/object.py
```diff
@@ -154,32 +139,147 @@ def nvds_generate_obj_uid(
...
+            'No rotated bounding box found in Deepstream meta information.'
+        )
+    rbbox.x_center = bbox.xc
+    rbbox.y_center = bbox.yc
+    rbbox.width = bbox.width
+    rbbox.height = bbox.height
+    rbbox.angle = bbox.angle
+
+
+def get_aligned_bbox_from_obj_meta(obj_meta: pyds.NvDsObjectMeta) -> BBox:
```
dorgun: Unnecessary function, it is used only once, you can add code to nvds_upd_obj_bbox. In any case only nvds_upd_obj_bbox should be used.

savant/deepstream/utils/object.py
```diff
@@ -154,32 +139,147 @@ def nvds_generate_obj_uid(
...
+    else:
+        raise MetaValueError(
+            'No valid bounding box found in Deepstream meta information.'
+        )
+    return BBox.ltwh(
+        nv_ds_bbox.left, nv_ds_bbox.top, nv_ds_bbox.width, nv_ds_bbox.height
+    )
+
+
+def get_rotated_bbox_from_obj_meta(obj_meta: pyds.NvDsObjectMeta) -> RBBox:
```
dorgun: Unnecessary function, it is used only once, you can add code to nvds_upd_obj_bbox. In any case only nvds_upd_obj_bbox should be used.

savant/deepstream/utils/object.py
```diff
@@ -154,32 +139,147 @@ def nvds_generate_obj_uid(
...
+
+    elif isinstance(bbox, RBBox):
+        set_rotated_bbox_for_obj_meta(batch_meta, obj_meta, bbox)
+        nvds_set_obj_selection_type(
+            obj_meta,
+            ObjectSelectionType.ROTATED_BBOX,
+        )
+
+
+def set_aligned_bbox_for_obj_meta(obj_meta: pyds.NvDsObjectMeta, bbox: BBox) -> None:
```
dorgun: Bring to a single notation. There are set_, get_, and nvds_set.

dorgun (changes requested):

savant/deepstream/drawfunc.py
abramov-oleg: Positional args are supposed to be more performant https://peps.python.org/pep-0570/#performance

When per-object functions are concerned it might make a difference.

Readability disadvantage seems to be minor and largely mitigated by IntelliSense-like IDE features.

savant/deepstream/pipeline.py
abramov-oleg: Removed

savant/deepstream/pipeline.py
abramov-oleg: Removed

savant/deepstream/utils/object.py
abramov-oleg: Avoiding code duplication is not the only reason to extract methods.

In this case I think it significantly improves readability.

savant/deepstream/utils/object.py
abramov-oleg: I think it improves readability.

savant/deepstream/utils/object.py
abramov-oleg: I think it improves readability.

savant/deepstream/utils/object.py
abramov-oleg: Don't understand the required changes.

Should I add `nvds_` prefix to all functions?

savant/deepstream/utils/object.py
dorgun: See above

savant/deepstream/utils/object.py
dorgun: See above

savant/deepstream/utils/object.py
dorgun: 1) Some functions start with "set", others with "update", but they do the same thing.
2) We gave names to all functions beginning with "nvds", which work with deepstream meta

savant/deepstream/utils/object.py
dorgun: For me personally, this only makes it worse to read, because it requires you to go into a function to understand what's going on, and the function just assigns values. Also, you replaced the passing of arguments by using positional arguments and you say that reducing readability is less important because it adds performance. And in this case you reduce performance, but say it increases readability. If we are looking for performance, then we should definitely get rid of functions consisting of several lines and only called in one place.

abramov-oleg: Refactored.

savant/deepstream/utils/object.py
abramov-oleg: Refactored.

dorgun (approved):

PR #301, opened 2023-07-19
#285 don't restore original resolution on module output

### LINKED ISSUE #285: Encoded video with padding breaks always-on-rtsp
The problem can be reproduced with the specified video [Towards_chengalpet.mp4](https://merlin-assets-bucket-dev.s3.ap-south-1.amazonaws.com/mwc-sample-videos/Towards_chengalpet.mp4).

The original video resolution is 3072x2048. If we define this resolution as a module frame parameters, the adapter will work with any paddings. If we define frame resolution 1280x720 (default) and set the top or bottom padding for the module, the adapter will raise the error 'nvbufsurface : Unsupported output format (36)' and we will need to restart it (the adapter won't break, but the output will just be a stub)

### Reproduce
Use any simple pipeline with frame padding top/bottom specified, eg.
```
name: ${oc.env:MODULE_NAME, 'test'}
parameters:
  frame:
    width: 1280
    height: 720
    padding:
      left: 0
      right: 0
      top: 0
      bottom: 100
  batch_size: 1
  output_frame:
    codec: h264
pipeline:
  source:
    element: uridecodebin
    properties:
      uri: file:///data/Towards_chengalpet.mp4
```

Run the module as usual, eg. `make run-dev`, then inside `python -m savant.entrypoint path/to/config.yml`

Run adapter, eg.
```
./scripts/run_sink.py always-on-rtsp --dev-mode --source-id=0 --stub-file-location=samples/assets/stub_imgs/smpte100_1280x720.jpeg --publish-ports
```

tomskikh:


savant/config/schema.py
```diff
@@ -55,6 +55,18 @@ def total_height(self) -> int:
             return self.height + self.padding.top + self.padding.bottom
         return self.height
 
+    @property
+    def output_width(self) -> int:
```
abramov-oleg, line 63: Let's add docstrings to `FrameParameters` properties.

abramov-oleg (changes requested):

abramov-oleg (approved):

PR #310, opened 2023-07-20
#298 implement uri-input.py script

### LINKED ISSUE #298: FFmpeg input script in Savant image
We must add a script `ffmpeg-input.py` resolvable through standard paths when the user `execs` into the container with a module. 

The script, by default, uses a standard socket pair and connection mode to ingest data in the framework. It also supports the optional argument `--socket=...`, allowing to override the type. Other arguments the script receives include:

- `-url=...`;
- `-source_id=` (default eq 'test');

tomskikh:


scripts/uri-input.py
```diff
@@ -0,0 +1,246 @@
...
+from typing import Any, Dict, Optional
+from urllib.parse import urlparse
+
+import click
+
+from savant.utils.file_types import FileType, parse_mime_types
+
+logger = logging.getLogger('uri-input')
+
+adapters_dir = Path(__file__).parent.parent / 'adapters' / 'gst' / 'sources'
```
abramov-oleg: Let's rename the `adapters_dir` variable using uppercase, to distinguish from local variables.

abramov-oleg (changes requested):

abramov-oleg (approved):

PR #318, opened 2023-07-25
#271 Implement software encoder (non NVENC) support

### LINKED ISSUE #271: Implement software encoder (non NVENC) support
When the user doesn't need to transcode all the streams (conditional processing), a software encoder can be efficiently used. It also benefits A100, Nano V2, and other GPUs with limited or missing NVENC.

tomskikh:


docs/source/savant_101/20_video_processing.rst
```diff
@@ -186,10 +186,13 @@ If the ``output_frame`` section is omitted, video frames will not be sent to sin
...
 
 - RAW RGBA (not optimal, as it requires large transfers over PCI-E);
-- JPEG (software);
-- PNG (software);
-- H264 (hardware);
-- HEVC (H265, hardware).
+- JPEG (hardware ``nvjpegenc``, software ``jpegenc``);
+- PNG (software ``pngenc``);
+- H264 (hardware ``nvv4l2h264enc``, software ``x264enc``);
+- HEVC (H265, hardware ``nvv4l2h265enc``).
```
dorgun:
```suggestion
- HEVC/H265 (hardware ``nvv4l2h265enc``).
```

savant/gstreamer/codecs.py
```diff
@@ -25,63 +25,105 @@ class CodecInfo:
...
-                else:
-                    encoder = self.encoder_elements[-1]
+                self._sw_encoder = None
 
-            self._encoder = encoder
+        if encoder_type is None:
+            return self._nv_encoder or self._sw_encoder
 
-        return self._encoder
+        if encoder_type == 'nvenc':
```
dorgun, line 67: Why is it called nvenc? Can we call it hardware? I think it's easier to understand and if we suddenly support some other hardware, we can choose based on the hardware instead of the user being forced to choose.

bwsw: @dorgun won't fix now, let it be 'nvenc'.

dorgun (approved):

PR #319, opened 2023-07-25
#311 handle recv timeout for REQ socket

### LINKED ISSUE #311: Handle correct termination when a 0mq REQ socket reaches timeout
When a socket reaches timeout, the module must be terminated correctly, not with the exception. Use a proper termination code like `1` for the exit function. Use a proper message like `Error: The REP socket hasn't responded in a configured timeframe (xxxx) ms. Exit (1).`

image

tomskikh:


savant/utils/sink_factories.py
```diff
@@ -99,6 +100,10 @@ def __init__(
         socket_type: str = SenderSocketTypes.PUB.name,
         bind: bool = True,
         send_hwm: int = Defaults.SEND_HWM,
+        receive_timeout: int = Defaults.SENDER_RECEIVE_TIMEOUT,
+        # TODO: Pipeline hangs on shutdown when zmq.Again raised.
```
dorgun: Not clear what this TODO is about? What does it mean to hang and change the value?

adapters/gst/gst_plugins/python/zeromq_sink.py
```diff
@@ -182,7 +211,25 @@ def do_render(self, buffer: Gst.Buffer):
         )
         self.sender.send_multipart(message)
         if self.wait_response:
-            resp = self.sender.recv()
+            try:
+                resp = receive_response(self.sender, self.req_recv_retries)
+            except zmq.Again:
+                error = (
+                    f"The REP socket hasn't responded in a configured timeframe "
+                    f"{self.receive_timeout * self.receive_timeout} ms."
```
dorgun: Why is receive_timeout squared?

savant/utils/sink_factories.py
tomskikh: When sink_factory raise an error `zmq.Again` pipeline hangs on shutdown. This can happen when socket is `req+connect`.

adapters/gst/gst_plugins/python/zeromq_sink.py
tomskikh: That's a typo, fixed it.

savant/utils/sink_factories.py
bwsw: @tomskikh how can we handle it to avoid the hang?

tomskikh: @bwsw I tried to add `exit(1)` on this error, but this didn't work.

```python
        with NvDsPipelineRunner(pipeline):
            try:
                for msg in pipeline.stream():
                    sink(msg, **dict(module_name=config.name))
            except Exception as exc:  # pylint: disable=broad-except
                logger.error('Error: %s',exc, exc_info=True)
                exit(1)
```

Maybe we need to drain the queue (`pipeline.stream()`) when shutting down the pipeline. I suggest to keep this workaround (`req_receive_retries: int =  1000`) for now and create a separate issue to fix this.

bwsw: Consider: 
https://stackoverflow.com/questions/905189/why-does-sys-exit-not-exit-when-called-inside-a-thread-in-python

tomskikh: Added `os._exit(1)`.

dorgun (approved):

PR #346, opened 2023-08-09
Set draw_func to occupy the pipeline before demuxer

### LINKED ISSUE #322: Set draw_func to occupy the pipeline before demuxer
Need to move it to be placed before demuxer it and queue before it.

tomskikh:


savant/deepstream/base_drawfunc.py
```diff
@@ -1,30 +1,54 @@
...
+
+    def can_draw_on_frame(self, frame_meta: NvDsFrameMeta) -> bool:
+        """Check whether we can draw on this specific frame."""
+
+        if self.condition.tag is None:
+            return True
+
+        if self.condition.tag in frame_meta.tags:
+            return True
+        else:
```
denisvmedyantsev: No need for `else` here.

savant/config/schema.py
```diff
@@ -232,8 +243,42 @@ def get_element_name(element: Union[DictConfig, PipelineElement]) -> str:
...
+            {
+                'module': self.module,
+                'class': self.class_name,
+                'kwargs': json.dumps(kwargs),
+            }
+        )
+
+
+@dataclass
+class DrawFunc(PyFuncElement):
```
denisvmedyantsev, line 280: What is the reason for defining DrawFunc as an element? Do we allow the user to use DrawFunc as a pipeline element and insert it on their own?

denisvmedyantsev:
I tried some of the samples (perf conf versions) and got into the log 
```GStreamer-CRITICAL **: 16:43:12.799: gst_element_release_request_pad: assertion 'GST_PAD_PAD_TEMPLATE (pad) == NULL || GST_PAD_TEMPLATE_PRESENCE (GST_PAD_PAD_TEMPLATE (pad)) == GST_PAD_REQUEST' failed```

I haven't found out if it affects the pipeline.

**Review by denisvmedyantsev — CHANGES_REQUESTED** [review]

File: savant/config/schema.py (hunk 2 above)
Comment (bwsw) [reply savant/config/schema.py:280]: @denisvmedyantsev 
I don't see the point as relevant. It can be an element if there are no certain problems with that. Because it is an element but always placed last. The user doesn't care, but from our perspective, we don't need various blocks to be represented by various concepts without the need. I don't see a need to make DrawFunc differ from PyFunc.

File: savant/deepstream/pipeline.py [hunk 3]
```diff
@@ -121,6 +132,67 @@ def __init__(
...
+                    element,
+                    queue_properties,
+                    last_is_queue,
+                    next_should_be_queue,
+                )
+                next_should_be_queue = False
+                elements.append(element)
+                continue
+
+            if (next_should_be_queue and element.element != 'queue') or (
```
denisvmedyantsev: I think you can also check if there are elements before/after pyfunc. You don't need to add a queue between muxer and pyfunc and pyfunc and demuxer.

savant/deepstream/drawfunc.py
```diff
@@ -41,20 +40,11 @@ def __init__(self, **kwargs):
             self.default_spec_track_id = get_default_draw_spec(track_id=True)
             self.default_spec_no_track_id = get_default_draw_spec(track_id=False)
 
-        self.frame_streams = []
-
-    def __call__(self, nvds_frame_meta: pyds.NvDsFrameMeta, buffer: Gst.Buffer):
```
denisvmedyantsev, line 46: If necessary, fix vlf-pipeline modules that use the old interface.

tomskikh: I've created MR in vlf-pipelines.

tomskikh:
> I tried some of the samples (perf conf versions) and got into the log `GStreamer-CRITICAL **: 16:43:12.799: gst_element_release_request_pad: assertion 'GST_PAD_PAD_TEMPLATE (pad) == NULL || GST_PAD_TEMPLATE_PRESENCE (GST_PAD_PAD_TEMPLATE (pad)) == GST_PAD_REQUEST' failed`
> 
> I haven't found out if it affects the pipeline.

I'm trying to release src pad instead of sink pad by mistake. Looks like we don't need to release pads at all.

savant/deepstream/pipeline.py
tomskikh: Fixed this.

savant/deepstream/base_drawfunc.py
tomskikh: Removed `else`.

denisvmedyantsev (approved):

PR #351, opened 2023-08-11
bug with incorrect parsing of ZMQ endpoint fixed

### LINKED ISSUE #350: Auxiliary scripts do not work with the short notation of the zmq endpoint
When using the run_sink, run_module, and run_source helper scripts, modules do not send or receive data through ZMQ

dorgun:


savant/utils/re_patterns.py
```diff
@@ -0,0 +1,4 @@
+import re
+
+socket_uri_pattern = re.compile('([a-z]+\\+[a-z]+:)?([a-z]+://.*)')
```
abramov-oleg, line 3: Since we expect only a very limited set of possible words, I'd write more concrete regex for the socket options

Something like

(sub|pub|req|rep|etc...)+(sub|pub|req|rep|etc...)

Up to you.

abramov-oleg (approved):

PR #359, opened 2023-08-16
Pyfunc dynamic reloading

### LINKED ISSUE #341: PyFunc/DrawFunc dynamic reloading
A developer needs instruments to develop interactively. However, GStreamer/DeepStream takes time to initialize, which makes the development less productive. To overcome that, we must introduce a new feature allowing the pipeline to reload Python functions specified in YAML and the draw_function to be reloaded when the changes are detected.

The function is optional and configured with a global parameter defined in YAML: `dynamic_reload: bool`. If it is set to `true` the feature is enabled; otherwise, the code doesn't reload on change. By default, the flag must be set to `false` which is overridable with an environment variable `DYNAMIC_RELOAD`.

Dynamic reloading works on a per-message basis (that is why it is important to make it in a way to avoid extra checks when `dynanic_reload` is set to `false`). The framework, when forwards a message to a particular pyfunc checks if the corresponding module is changed. If there is a change, it reloads the module and recreates the Object according to the manifest (kwargs, yada-yada). The state is not copied from the old object. Finally, it replaces the old object instance with a new one and forwards the traffic.

It is important to implement an ultimate exception-handling solution covering all unhandled exceptions and pretty stack trace presentation (e.g., with https://pypi.org/project/pretty-traceback/).

When reloading happens, it must inform the user about it:

```
INFO Dynamic reloading is enabled and changes in `module.name.Cls` detected; reloading.
INFO The module `module.name.Cls` reloading complete: OK
OR
ERROR The module `module.name.Cls` reloading failed.
ERROR ------------------------ BEGIN ------------------------------------
ERROR stacktrace ... (pretty stacktrace)
ERROR ------------------------- END -------------------------------------
```

Keep in mind there could be multiple classes declared in a single module, so the reloading can be suboptimal from the module perspective (multiple reloads).

Please investigate if GST output messages also can be colored to conform the whole situation.

abramov-oleg:


savant/utils/logging/savant_rs_handler.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def emit(self, record: logging.LogRecord):
+        """Emit a record."""
+        try:
+            formatted = self.format(record)
+        except Exception:
+            self.handleError(record)
+            return
+
+        record_name_rs = record.name.replace('.', '::')
```
bwsw: @abramov-oleg 
You can remove that. In `0.1.60` it is implemented internally.

abramov-oleg: Updated to 0.1.60 and removed this replace call.

adapters/ds/sinks/always_on_rtsp/__main__.py
```diff
@@ -18,12 +18,12 @@
 from savant.gstreamer.codecs import CODEC_BY_CAPS_NAME, Codec
 from savant.gstreamer.element_factory import GstElementFactory
 from savant.gstreamer.runner import GstPipelineRunner
-from savant.utils.logging import get_logger
 from savant.utils.platform import is_aarch64
 from savant.utils.zeromq import ReceiverSocketTypes
+from savant.utils.logging import init_logging
```
tomskikh: Imports are not sorted and `make run-isort` does nothing. Does it work?

savant/config/module_config.py
```diff
@@ -207,6 +223,9 @@ def apply_schema(
         OmegaConf.structured(FrameParameters),
     )
     apply_schema(module_cfg.parameters, 'draw_func', DrawFunc)
+    if module_cfg.parameters.dev_mode:
+        logger.debug('Setting draw_func dev mode to true.')
+        module_cfg.parameters.draw_func.dev_mode = True
```
tomskikh, line 227: Handle the case when draw_func is disabled

savant/utils/inotify_manager.py
```diff
@@ -0,0 +1,85 @@
+"""Inotify manager module."""
+from collections import defaultdict
+import logging
+from inotify_simple import INotify, flags
+from savant.utils.singleton import SingletonMeta
+
+
+class INotifyManager(metaclass=SingletonMeta):
```
tomskikh, line 10: Why is it a singleton and not just a regular class for a single file? Looks like an unnecessary complication.

savant/utils/logging/log_setup.py
```diff
@@ -0,0 +1,99 @@
...
+
+def get_default_loglevel() -> str:
+    return os.environ.get('LOGLEVEL', 'INFO')
+
+
+def get_log_conf(log_level: str) -> dict:
+    return {
+        'version': 1,
+        'formatters': {
+            # 'basic': {},
```
tomskikh: Remove commented lines or add a comment why do we need them here.

savant/utils/logging/log_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    finally:
+        logging._releaseLock()
+
+
+def log_level_py_to_rs(py_log_level: int):
+    """Get a rust log level corresponding to the specified python log level.
+
+    :param py_log_level: python log level as an int.
+    """
+    if py_log_level in (logging.ERROR, logging.CRITICAL):
```
tomskikh: I think better to make it a dict.

savant/base/pyfunc.py
```diff
@@ -109,53 +140,239 @@ class PyFunc:
...
+                logger.debug('No module spec, setting watch is impossible.')
+            elif not spec.has_location:
+                logger.debug(
+                    'Module spec with undetermined location, setting watch is impossible.'
+                )
+            else:
+                logger.debug('Setting watch for pyfunc module source.')
+                INotifyManager().add_watch(spec.origin, id(self))
+
+        # try to instatiate module
```
tomskikh:
```suggestion
        # try to instantiate module
```

savant/base/pyfunc.py
```diff
@@ -109,53 +140,239 @@ class PyFunc:
...
+                    logger.exception('Error calling on_start()')
+                else:
+                    logger.info(
+                        'The module "%s.%s" reloading complete: OK',
+                        self.module,
+                        self.class_name,
+                    )
+            else:
+                logger.info(
+                    'The module "%s.%s" reloading finsihed: Fail',
```
tomskikh:
```suggestion
                    'The module "%s.%s" reloading finished: Fail',
```

savant/config/default.yml
```diff
@@ -114,6 +114,11 @@ parameters:
     # Parameters for telemetry provider
     provider_params: ${json:${oc.env:TELEMETRY_PROVIDER_PARAMS, null}}
 
+  # Flag controls whether user python modules (pyfuncs and drawfuncs)
+  # are monitored for changes during runtime
+  # and reloaded in case changes are detected
+  dev_mode: ${oc.decode:${oc.env:DEV_MODE, False}}
```
tomskikh, line 120: Update line numbers in `literalinclude` directives in `docs/`.

tomskikh (changes requested):

savant/utils/logging/savant_rs_handler.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Hardcoded to use two formatters (no need to specify during logging configuration):
+    1. basic formatter for all messages under ERROR priority
+    2. pretty_traceback formatter for ERROR and CRITICAL messages
+    """
+
+    def __init__(self) -> None:
+        logging.Handler.__init__(self)
+        self.pretty_trace_back_threshold = logging.ERROR
+        self.formatter_basic = logging.Formatter()
```
tomskikh: Add timestamps to log formatter
```
DEBUG savant::pipeline::demo                                 > Requesting new sink pad on muxer: sink_0
INFO  savant::pipeline::demo                                 > Added source cam-1
```

abramov-oleg: The idea is that the underlying rust logger (currently [pretty_env_logger](https://github.com/insight-platform/savant-rs/blob/09c5533740de5267912df6fba3556dfa1a0219cf/savant_python/src/lib.rs#L65)) will do the formatting.

bwsw: @abramov-oleg you are right

adapters/ds/sinks/always_on_rtsp/__main__.py
abramov-oleg: Merged with latest isort fix and ran reformat

savant/config/module_config.py
abramov-oleg: Added a check

savant/utils/logging/log_utils.py
abramov-oleg: Changed to set

savant/utils/logging/log_setup.py
abramov-oleg: Removed

savant/utils/logging/log_utils.py
tomskikh: No, I mean use a dict `py_log_level -> savant_rs_log_level` instead of if-else.

savant/utils/inotify_manager.py
abramov-oleg: It seems the best practice is not to use more than a single instance of inotify (for example, [relevant discussion](https://github.com/sublimehq/sublime_text/issues/1195))

It seemed convenient to manage access to this instance and to the single (combined) event queue for all the watched files through a singleton.

Can you elaborate how I can simplify this?

tomskikh: Didn't know about this issue. In that case a singleton makes sense.

tomskikh (approved):

Issue #353, opened 2023-08-11
Refactor python imports


denisvmedyantsev:
Lack of order in imports leads to duplicates and makes it harder to understand the code (eg. https://github.com/insight-platform/Savant/blob/develop/savant/deepstream/pipeline.py, https://github.com/insight-platform/Savant/blob/develop/savant/deepstream/buffer_processor.py). 

Consider adding a utility that will organize import statements.

PR #361, opened 2023-08-17
Refactor python imports (#353)

FrankXenarthra:
Lack of order in imports leads to duplicates and makes it harder to understand thecode.

The issue only refereced two files but with the isort utility every python file has been modified automatically.


denisvmedyantsev:
@FrankXenarthra Thanks for your PR.

We use [black](https://github.com/psf/black) in the project (https://github.com/insight-platform/Savant/blob/develop/Makefile#L131) and the new multi-line import format does not satisfy their rules. The multi-line import format needs to be corrected.

What tool did you use to fix the imports? We need a tool to manually reformat imports via make (like black).

bwsw:
@FrankXenarthra @denisvmedyantsev looks like isort supports black:

https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

FrankXenarthra:
@bwsw @denisvmedyantsev I would create a new PR including an .isort.cfg with black profile and apply isort then for the commit. Do you agree? Is it the correct approach to always create a new PR in case I ammended a commit? Thanks for the clarification!

bwsw:
@FrankXenarthra add a commit right to the current PR, no need creating new PR.

denisvmedyantsev:
I missed the mention of isort. @bwsw, you are right, we only need the section in the pyproject.toml file
```
[tool.isort]
profile = "black"
```
Then relaunch isort.

denisvmedyantsev:
@FrankXenarthra, one more thing. Please, add isort to `requirements/dev.txt` and commands check/run-isort in the `Makefile`.

FrankXenarthra:
@bwsw @denisvmedyantsev I implemented the propose changes. Couldn't you have a look please?

denisvmedyantsev (approved):

PR #381, opened 2023-08-28
Enforce frame geometry divisible on 8

### LINKED ISSUE #370: Enforce frame geometry divisible on 8
- scaled size
- paddings

tomskikh:


denisvmedyantsev:
It's not a parameter to set, it's an internal constraint (nvstreammux). Isn't it? @bwsw

bwsw:
@denisvmedyantsev yes, but we don't know how it can change between devices. So I suggested it would be better to not hardcode it.

denisvmedyantsev (approved):

PR #382, opened 2023-08-28
Support Ignorable shutdown message

### LINKED ISSUE #371: Support Ignorable shutdown message
We need to support a global shutdown message causing the pipeline to shut down. The message contains a key that must match to affect the shutdown. The pipeline can be configured to respect or ignore shutdown messages.

tomskikh:


gst_plugins/python/savant_rs_video_demux.py
```diff
@@ -408,6 +429,25 @@ def handle_eos(self, eos: EndOfStream) -> Gst.FlowReturn:
 
         return Gst.FlowReturn.OK
 
+    def handle_shutdown(self, shutdown: Shutdown) -> Gst.FlowReturn:
+        """Handle Shutdown message."""
+        if self.shutdown_auth is None or shutdown.auth != self.shutdown_auth:
+            self.logger.debug('Ignoring shutdown message.')
+            return Gst.FlowReturn.OK
+
+        self.logger.info('Received shutdown message.')
```
abramov-oleg: Let's make logging a bit more detailed here by adding the reason the shutdown message was ignored.

abramov-oleg (approved):

PR #392, opened 2023-09-01
391 documentation update

### LINKED ISSUE #391: Ongoing documentation update

bwsw:
Closes #391


docs/source/advanced_topics/0_batching.rst
```diff
@@ -1,22 +1,34 @@
...
 
-Increasing the batch size may not improve performance depending on the primary model. In this case, you may stay safe using the minimal batch size, which results in optimal performance (not necessary to increase it to the theoretical number of parallel streams processed). E.g., you expect the pipeline to handle up to 16 streams; you also discovered that the model saturates its performance with a batch size of 4. Therefore, you should use four as a pipeline batch size.
+Stream Multiplexing Batching
+----------------------------
 
-The pipeline configuration file has parameters that regulate the batch size. They are ``batch_size`` and ``batched_push_timeout``.
+In Savant, a batch consists of the frames from streams processed. It is important to note that the batch can include multiple frames from a single stream, so if a pipeline processes N streams, then the batch size can be from 1 to N (ideally N), but not necessarily all streams will be presented in every batch; also, a batch is not always full: when streams delay the frames, the batch can be gathered partially, decreasing the efficiency of the processing. Thus, usually, you want to maximize batch filling.
```
abramov-oleg: > when streams delay the frames, the batch can be gathered partially,

Suggest

> when streams delay the frames, the batch may be gathered partially,

docs/source/advanced_topics/0_batching.rst
```diff
@@ -1,22 +1,34 @@
...
-The ``batched_push_timeout`` defines how long the pipeline waits until the batch is full. The smaller value decreases latency, while the larger one results in greater processing speed. Typically, when handling the real-time streams at 30 FPS, you need to set ``batched_push_timeout`` to ``35000-40000`` (``35-40`` ms). However, when injecting video from file sources, keep it as low as ``1000000`` (``1`` ms).
+.. tip::
+
+    E.g., you expect the pipeline to handle up to 16 streams; you also discovered that the model saturates its performance with a batch size of 4. Therefore, you should use four as a pipeline batch size.
+
+The pipeline configuration file defines parameters regulating the batch size. They are ``batch_size`` and ``batched_push_timeout``.
+
+The ``batch_size`` parameter is configured for video multiplexing. When it is **not** configured, it is set based on the value defined for the first pipeline model. The default value is ``1``.
+
+The ``batched_push_timeout`` defines how long the pipeline waits until the batch is full. The smaller value decreases latency, while the larger one results in greater processing speed. The default value is: ``4000000`` (``4`` seconds).
```
abramov-oleg: > The ``batched_push_timeout`` defines how long the pipeline waits until the batch is full.

Suggest

> The ``batched_push_timeout`` defines the maximum time the pipeline waits for the batch to be fully formed, pushing out partially formed batches if the wait time exceeds this threshold.

abramov-oleg: > The smaller value decreases latency, while the larger one results in greater processing speed.

Suggest

> The smaller value decreases latency, while the larger one results in greater throughput.

docs/source/advanced_topics/0_batching.rst
```diff
@@ -26,10 +38,12 @@ Default value is: ``4000000`` (``4`` seconds).
...
 
-Secondary Batching
-------------------
+Model Batching
+--------------
+
+Models operate on objects detected by previous models or defined with ROIs (externally defined pseudo-objects). Usually, there are several objects in every frame, and there is a batch of frames. Thus, predicting how many objects you get after the previous model may be difficult.
 
-Secondary models operate on objects detected by primary models. Usually, there are several objects in every frame, and there is a batch of frames. Thus, predicting how many objects you get after the primary model may be difficult. However, the rule of thumb is to select the batch size typically set to a number between one and an average number of objects in the frame. You may experiment with various batch sizes to tweak the performance.
+However, the rule of thumb is to select the batch size typically set to a number between one and an average number of objects in the frame. You may experiment with various batch sizes to tweak the performance.
```
abramov-oleg:
```suggestion
However, the rule of thumb is to set the batch size to a number between one and an average number of objects in the frame. You may experiment with various batch sizes to tweak the performance.
```

docs/source/advanced_topics/2_element_group.rst
```diff
@@ -0,0 +1,28 @@
+Element Groups
+==============
+
+A single module config can contain multiple variations of the pipeline through the use of conditionally initialized :py:class:`~savant.config.schema.ElementGroup`.
+
+For example, the user may introduce alternative detector units without the need to duplicate all the other module parameters in a separate config file, as is done in the `traffic meter <https://github.com/insight-platform/Savant/tree/develop/samples/traffic_meter>`__ sample.
```
abramov-oleg:
```suggestion
For example, the user can introduce alternative detector units without the need to duplicate all the other module parameters in a separate config file, as is done in the `traffic meter <https://github.com/insight-platform/Savant/tree/develop/samples/traffic_meter>`__ sample.
```

docs/source/advanced_topics/9_open_telemetry.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Why To Use OpenTelemetry
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In streaming systems, data flows through container stages in parallel what leads to messy logs with overlapping records for various messages. For example, when the message ``M[N]`` enters the pipeline, message ``M[N-P]`` passes the stage ``P``, and their logs overlap.
+
+For a developer, navigating through those logs without advanced filtering tools is very difficult. OpenTelemetry solves the problem by introducing the concept of Trace: a unique ID corresponding to a business transaction. In Savant, every frame coming in a pipeline may have a trace id, thus making it possible to separate logs related to a specific frame.
+
+More to say, the Trace is not a flat structure: a developer can wrap certain pieces of code with so called Spans, limiting the scope, so logs are attached to the hierarchy of spans rather than Trace directly. Every span is automatically a profiling object because OpenTelemetry collects its start time, end time and duration.
```
abramov-oleg:
```suggestion
What's more, the Trace is not a flat structure: a developer can wrap certain pieces of code with so called Spans, limiting the scope, so logs are attached to the hierarchy of spans rather than Trace directly. Every span is automatically a profiling object because OpenTelemetry collects its start time, end time and duration.
```

docs/source/advanced_topics/9_open_telemetry.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Developers can attach to span auxiliary information: attributes, events, span statuses.
```
abramov-oleg:
```suggestion
Developers can attach auxiliary information to a span: attributes, events, span statuses.
```

docs/source/advanced_topics/9_open_telemetry.rst
```diff
@@ -0,0 +1,78 @@
...
+          -p 16686:16686 \
+          -p 4317:4317 \
+          -p 4318:4318 \
+          -p 14250:14250 \
+          -p 14268:14268 \
+          -p 14269:14269 \
+          -p 9411:9411 \
+          jaegertracing/all-in-one:1.48
+
+**TODO**: do we need to link the dev container with Jaeger container somehow, or it is available?
```
abramov-oleg: In general yes, the Jeager service needs to be accessible.

I've added 2 possible ways to connect module and client containers with the jaeger container to the template sample. Both are based on the named network configured in the docker compose file.

1. Run everything through docker compose.
2. Run only jaeger through docker compose, then run a dev container attached to the same named network.

Both ways allow to connect to jaeger using service name as host name.

Additionally, it's possible to start jaeger container as described in the comment above, but then add the  `--network host` flag to the module/client containers. Then the jaeger will be accessible on localhost

docs/source/getting_started/2_module_devguide.rst
```diff
@@ -0,0 +1,164 @@
...
+
+- **.devcontainer**: the directory is required only if you are developing with VSCode; it contains Docker specification for VSCode. You may find two flavors there - for x86+dGPU and for Nvidia Jetson SOCs.
+- **docker**: the directory contains two Dockerfiles containing stub code for building custom development Docker containers, installing additional requirements from the ``requirements.txt``; it is often a way to go if the pipeline requires Python libraries not delivered with Savant base images.
+- **.dockerignore**: the file contains exclusions for files and paths typically must be avoided to add to a Docker image; you can extend them to reflect your actual needs, e.g. exclude media files.
+- **Makefile** contains commands for building and launching the Docker image in a standalone mode, without IDE.
+- **README.md** is just a README stub.
+- **requirements.txt** contains additional Python requirements to add to the custom Docker image.
+- **module** directory with actual pipeline code.
```
abramov-oleg: I've placed the `module` dir on one level lower `src/module`, in order to mount client + module code into dev container easier.

docs/source/savant_101/00_streaming_model.rst
```diff
@@ -23,9 +23,9 @@ The module de-multiplexes those streams internally to the form where each stream
...
 
-Such a multiplexed/de-multiplexed operation model is very beneficial because any well-made Savant module immediately becomes production-ready. The module doesn't know whether it processes a file-based stream, a live stream, or just a bunch of images; it doesn't distinguish between them. The framework ensures the processing for all those streams in the same way. More to say, a mix of various streams can be processed simultaneously by the same module.
+Such a multiplexed/de-multiplexed operation model is very beneficial because any well-made Savant module immediately becomes production-ready. The module does not know whether it processes a file-based stream, a live stream, or just a bunch of images; it does not distinguish between them. The framework ensures the processing for all those streams in the same way. More to say, a mix of various streams can be processed simultaneously by the same module.
```
abramov-oleg:
```suggestion
Such a multiplexed/de-multiplexed operation model is very beneficial because any well-made Savant module immediately becomes production-ready. The module does not know whether it processes a file-based stream, a live stream, or just a bunch of images; it does not distinguish between them. The framework ensures the processing for all those streams in the same way. What's more, a mix of various streams can be processed simultaneously by the same module.
```

docs/source/advanced_topics/0_batching.rst
```diff
@@ -1,22 +1,34 @@
...
 
-Increasing the batch size may not improve performance depending on the primary model. In this case, you may stay safe using the minimal batch size, which results in optimal performance (not necessary to increase it to the theoretical number of parallel streams processed). E.g., you expect the pipeline to handle up to 16 streams; you also discovered that the model saturates its performance with a batch size of 4. Therefore, you should use four as a pipeline batch size.
+Stream Multiplexing Batching
+----------------------------
 
-The pipeline configuration file has parameters that regulate the batch size. They are ``batch_size`` and ``batched_push_timeout``.
+In Savant, a batch consists of the frames from streams processed. It is important to note that the batch can include multiple frames from a single stream, so if a pipeline processes N streams, then the batch size can be from 1 to N (ideally N), but not necessarily all streams will be presented in every batch; also, a batch is not always full: when streams delay the frames, the batch may be gathered partially, decreasing the efficiency of the processing. Thus, usually, you want to maximize batch filling.
```
abramov-oleg:
```suggestion
In Savant, a batch consists of the frames from streams processed. It is important to note that the batch can include multiple frames from a single stream, so if a pipeline processes N streams, then the batch size can be from 1 to N (ideally N), but not necessarily all streams will be present in every batch; also, a batch is not always full: when streams delay the frames, the batch may be gathered partially, decreasing the efficiency of the processing. Thus, usually, you want to maximize batch filling.
```

docs/source/advanced_topics/0_pipeline_benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+- non-real-time single stream performance;
+- non-real-time multiple stream performance;
+- real-time multiple stream performance.
+
+There is one more dimension to benchmark:
+
+- with adapters (end-to-end);
+- without adapters (only pipeline).
+
+When benchmarking, we usually measure the aggregate throughput of the pipeline in FPS (frames-per-second) it can handle. Sometimes, usually, when the pipeline is intended to work in real-time, we also measure end-to-end latency, but this document does not discuss it.
```
abramov-oleg:
```suggestion
When benchmarking, we usually measure the aggregate throughput of the pipeline in FPS (frames-per-second). When the pipeline is intended to work in real-time and in some other cases, we may also measure end-to-end latency, but this document does not discuss it.
```

docs/source/advanced_topics/0_pipeline_benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+        - element: devnull_sink
+
+At the end of operation, you will see the FPS result.
+
+Measuring Non-Real-Time Multiple Stream Performance
+---------------------------------------------------
+
+This kind of benchmarking is valuable to discover the maximum aggregate number of FPS the pipeline can reach. It may occur that the maximum value will be reached when ingesting 32 pipelines, but each of them will be processed at the rate of 5 FPS. That is why it is not real-time performance.
+
+To run such benchmarks we implemented a special adapter: :ref:`Multi-Stream Source Adapter <multi_stream_source_adapter>`, allowing the ingesting of a selected video file in parallel in the pipeline under benchmarking. By changing the number of parallel streams you may find out the value which is the maximum for the pipeline.
```
abramov-oleg:
```suggestion
To run such benchmarks we implemented a special adapter: :ref:`Multi-Stream Source Adapter <multi_stream_source_adapter>`, allowing the ingesting of a selected video file in parallel in the pipeline under benchmarking. By changing the number of parallel streams you can find out the maximum value for the pipeline.
```

docs/source/advanced_topics/0_pipeline_benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+.. code-block::
+
+    pipeline:
+      # noop pipeline sink, not using sink adapter
+      sink:
+        - element: devnull_sink
+
+
+It represents benchmarking without real-life sinks which can form a bottleneck. If you want to test CV/ML performance it is what you are looking for. However, often you need to test end-to-end, including a specific sink implementation used practically. In such a situation, you need to include in the benchmark additional components, like a sink adapter you are planning to use and 3rd-party systems.
```
abramov-oleg:
```suggestion
It represents benchmarking without real-life sinks which can form a bottleneck. This is what you are looking for if you want to test CV/ML performance . However, often you need to test end-to-end, including a specific sink implementation used practically. In such a situation, you need to include additional components in the benchmark, like a sink adapter you are planning to use and 3rd-party systems.
```

docs/source/advanced_topics/9_open_telemetry.rst
```diff
@@ -0,0 +1,82 @@
...
+.. tip::
+
+    Read an introductory `article <https://blog.savant-ai.io/opentelemetry-in-savant-instrumenting-deep-learning-computer-vision-pipelines-dd42c7a65d00?source=friends_link&sk=b5a0c1d8a3554a38862f0c35007c3452>`_ on OpenTelemetry in Savant on Medium.
+
+Why To Use OpenTelemetry
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In streaming systems, data flows through container stages in parallel what leads to messy logs with overlapping records for various messages. For example, when the message ``M[N]`` enters the pipeline, message ``M[N-P]`` passes the stage ``P``, and their logs overlap.
+
+For a developer, navigating through those logs without advanced filtering tools is very difficult. OpenTelemetry solves the problem by introducing the concept of Trace: a unique ID corresponding to a business transaction. In Savant, every frame coming in a pipeline may have a trace id, thus making it possible to separate logs related to a specific frame.
```
abramov-oleg:
```suggestion
For a developer, navigating through those logs without advanced filtering tools is very difficult. OpenTelemetry solves the problem by introducing the concept of Trace: a unique ID corresponding to a business transaction. In Savant, every frame coming in a pipeline can have a trace id, thus making it possible to separate logs related to a specific frame.
```

docs/source/savant_101/12_module_definition.rst
```diff
@@ -110,7 +114,27 @@ In the above case, environment variables have lesser priority than values fetche
 Output Queue Max Size
 ^^^^^^^^^^^^^^^^^^^^^
 
-TODO: queue_maxsize
+The ``queue_maxsize`` parameter specifies the size of the buffer located at the end of the pipeline, right before the ZeroMQ sink. The parameter may be beneficial in cases when metadata payload metadata fluctuates between sequential frames in size significantly, causing temporary output delays. The buffer helps avoid blocking the pipeline while it has spare capacity. The default value is ``100``.
```
abramov-oleg:
```suggestion
The ``queue_maxsize`` parameter specifies the size of the buffer located at the end of the pipeline, right before the ZeroMQ sink. The parameter may be beneficial in cases when payload metadata fluctuates between sequential frames in size significantly, causing temporary output delays. The buffer helps avoid blocking the pipeline while it has spare capacity. The default value is ``100``.
```

docs/source/savant_101/12_module_definition.rst
```diff
@@ -110,7 +114,27 @@ In the above case, environment variables have lesser priority than values fetche
...
+.. warning::
+
+    Large values cause extra GPU/CPU memory usage.
+
+.. _buffering_queues:
+
+Buffering Queues
+^^^^^^^^^^^^^^^^
+
+The ``buffer_queues`` parameter is used to enable Python parallelization and enchance the performance in compute-intensive pipelines. By default, the parameter is disabled.
```
abramov-oleg:
```suggestion
The ``buffer_queues`` parameter is used to enable Python parallelization and enhance the performance in compute-intensive pipelines. By default, the parameter is disabled.
```

abramov-oleg (changes requested):

abramov-oleg (approved):

PR #394, opened 2023-09-04
Fix single label multi obj class mapping

### LINKED ISSUE #386: Fix how the mapping to a single label works
Currently it works incorrectly, just losing the information.

abramov-oleg:


savant/deepstream/nvinfer/element_config.py
```diff
@@ -32,7 +33,9 @@
 from savant.remote_file import process_remote
 from savant.utils.logging import get_logger
 
-__all__ = ['nvinfer_element_configurator']
+__all__ = ['nvinfer_element_configurator', 'MERGED_CLASSES']
+
+MERGED_CLASSES = defaultdict(dict)
```
bwsw, line 38: @abramov-oleg maybe use per-model scopped mapping, not global? It may introduce collisions.

abramov-oleg: Yes, I already use a two-level dict with the first level being tied to the element name. So, for example, multiple detector units based on the same detector model will not have any configuration collisions.

savant/deepstream/nvinfer/element_config.py
```diff
@@ -411,14 +414,36 @@ def process(self, msg, kwargs):
...
-        object_labels = {
-            obj.class_id: obj.label for obj in model_config.output.get('objects')
-        }
-    model_uid = register_model_objects(
-        element_config.name, object_labels, RegistrationPolicy.Override
-    )
+    if issubclass(model_type, (ObjectModel, ComplexModel)):
+        output_objects = model_config.output.get('objects', [])
+        if output_objects:
+            logger.debug('Registering output objects for the model ...')
```
denisvmedyantsev, line 420: Try to name the model here, in logger.debug.

savant/deepstream/buffer_processor.py
```diff
@@ -694,6 +695,19 @@ def prepare_element_output(self, element: PipelineElement, buffer: Gst.Buffer):
                                 obj_label = build_model_object_key(
                                     element.name, obj.label
                                 )
+                                obj_cls_id = MERGED_CLASSES[element.name].get(
```
denisvmedyantsev, line 698: Are you sure there is no option for element.name not to be represented in MERGED_CLASSES?

denisvmedyantsev (approved):

savant/deepstream/buffer_processor.py
abramov-oleg: Should be ok, as the `MERGED_CLASSES` is a defaultdict

savant/deepstream/nvinfer/element_config.py
abramov-oleg: It's already included in all messages in this function through the logging adapter defined above.

PR #397, opened 2023-09-05
Expose Savant health status to HTTP

### LINKED ISSUE #395: Expose Savant health status to HTTP
We must extend the health check functionality to expose it as an HTTP endpoint for external clients. Specifically, we need a feature in Client SDK (a.k.a. Python Source/Sink SDK) to be able to wait until the module is up.

In the module, we may implement it as a simple HTTP server serving static uncached or with Nginx.

The most straightforward approach is implementing it in a separate Python thread with https://docs.python.org/3/library/http.server.html to avoid using Supervisord.

From the Client SDK perspective, I propose an addition to SourceBuilder and SinkBuilder, an optional build arg specifying the health check and timeout to wait until it is healthy before the source or sink starts operating.

E.g.

```python
source = (
        SourceBuilder()
        .with_log_provider(JaegerLogProvider('http://localhost:16686'))
        .with_socket('pub+connect:ipc:///tmp/zmq-sockets/input-video.ipc')
        .with_module_health_check('http://1.1.1.1:8888/healthcheck')          # feature
        .build()
    )
```

SinkBuilder receives a similar property.

tomskikh:


savant/client/builder/sink.py
```diff
@@ -81,6 +116,8 @@ def _with_field(self, field: str, value) -> 'SinkBuilder':
                 'socket': self._socket,
                 'log_provider': self._log_provider,
                 'idle_timeout': self._idle_timeout,
+                'module_health_check_timeout': self._module_health_check_timeout,
```
dorgun, line 120:
```suggestion
                'module_health_check_url': self._module_health_check_url,
                'module_health_check_timeout': self._module_health_check_timeout,
```

savant/entrypoint/main.py
```diff
@@ -36,6 +38,21 @@ def main(config_file_path: str):
...
+                host='',
+                port=healthcheck_port,
+                http_path='/healthcheck',
+                status_filepath=status_filepath,
+            )
+            healthcheck_thread = Thread(
+                target=healthcheck_server.serve_forever,
+                daemon=True,
+            )
+            healthcheck_thread.start()
```
dorgun, line 54: Wouldn't there be some problem with the thread not stopping when the pipelines are terminated?

dorgun (changes requested):

savant/entrypoint/main.py
tomskikh: No, this is a daemon thread.

dorgun (approved):

PR #411, opened 2023-09-12
Update perf benchmark

### LINKED ISSUE #352: Run benchmarks with queues and without them
No queues, q=1, q=10
x86, Jetson NX

denisvmedyantsev:


savant/config/json_resolver.py
```diff
@@ -16,7 +16,10 @@ def json_resolver(  # pylint:disable=unused-argument
...
     logger.debug('Parsing param JSON "%s"', json_string)
     try:
-        parsed_conf_node = json.loads(json_string)
+        if json_string is not None:
+            parsed_conf_node = json.loads(json_string)
+        else:
+            return None
     except TypeError:
         logger.warning('JSON loads fail, returning None for "%s".', json_string)
         return None
```
abramov-oleg: Is this to avoid seeing the warning when json str is None?

I think the following is a bit more readable, but up to you.

```python

try:
    parsed_conf_node = json.loads(json_string)
except TypeError:
    if json_string:
        logger.warning('JSON loads fail, returning None for "%s".', json_string)
    return None
```

savant/utils/artist/artist_gpumat.py
```diff
@@ -147,6 +147,10 @@ def add_bbox(
             left, top, right, bottom = bbox.visual_box(
                 PaddingDraw(*padding), border_width, self.max_col, self.max_row
             ).as_ltrb_int()
+
+            if (right - left) == 0 or (bottom - top) == 0:
+                raise ValueError('Wrong bbox size.')
```
abramov-oleg: Maybe make the check like so

```suggestion
            if (right - left) < 1 or (bottom - top) < 1:
                raise ValueError('Wrong bbox size.')
```

since we can't rely on `bbox.visual_box()` to return sane values.

Same for the width and height check in `blur()`.

scripts/run_module.py
```diff
@@ -105,21 +116,26 @@ def run_module(
         )
 
     module_name = module_config_path.parent.stem
-    volumes = [f'{(repo_root_dir / "samples").resolve()}:/opt/savant/samples']
+    if module_name == 'module':
+        module_name = module_config_path.parent.parent.stem
```
abramov-oleg: Is this to account for dir structure as in `yolov8seg` sample?

`face_reid` is left unsupported currently.

Let's make it more generalized: go up in dir tree until `samples` is found, then previous dir name is the module name.

scripts/run_perf.py
```diff
@@ -21,75 +17,268 @@
...
-            if process.stderr:
-                yield process.stderr.readline().strip()
-            if ret_code is not None:
-                break
+    # remove perhaps outliers
+    if len(fps_list) > 2:
+        fps_list = fps_list[1:-1]
+    # average
+    if len(fps_list) > 1:
+        num_frames = len(fps_list) * FPS_PERIOD
```
abramov-oleg: Won't this produce incorrect `num_frames` in case the number of frames in the source isn't evenly divisible by the FPS_PERIOD?

Will it be better to get num frames directly from fps messages by capturing the `\d+` in `Processed \d+ frames`?

abramov-oleg (changes requested):

savant/deepstream/drawfunc.py
```diff
@@ -185,4 +194,14 @@ def _draw_central_dot(self, obj_meta: ObjectMeta, artist: Artist, spec: DotDraw)
...
+            artist.blur(obj_meta.bbox)
+        except Exception as exc:
+            self.logger.warning(
+                'Got "%s: %s" when trying to blur the object %s.',
+                type(exc).__name__,
+                exc,
+                obj_meta.bbox,
+                exc_info=exc,
+            )
+
```
abramov-oleg: Run `make reformat` to remove empty line.

savant/config/json_resolver.py
denisvmedyantsev: Your code is more concise, but I think it's better to check than to raise and catch an exception. I'll rewrite the code a little bit.

scripts/run_module.py
denisvmedyantsev: I will rewrite this piece of code. But it still won't allow you to run the `face_reid` module, because you used a specific volume mapping in the `face_reid` module.

savant/utils/artist/artist_gpumat.py
denisvmedyantsev: ok

scripts/run_perf.py
denisvmedyantsev: You are right. I'll fix it.

savant/deepstream/drawfunc.py
denisvmedyantsev: ok

abramov-oleg (approved):

PR #436, opened 2023-09-21
Unify python and rust logging config

### LINKED ISSUE #428: The module requires two envs to set DEBUG log level
To set log level to DEBUG we need to specify two envs: `LOG_LEVEL` and `RUST_LOG`. Fix this to use only `LOG_LEVEL`.

abramov-oleg:


savant/utils/logging/log_utils.py
```diff
@@ -1,18 +1,116 @@
...
+    'trace': LogLevel.Trace,
 }
 
 
+def parse_log_spec(log_spec_str: str) -> dict:
+    """Parse logging specification string.
+
+    :param log_spec_str: A comma-separated list of logging directives
+        of the form target=level.
+    :return:  A dictionary of of the form log_target:log_level.
```
tomskikh:
```suggestion
    :return:  A dictionary of the form log_target:log_level.
```

savant/utils/logging/log_utils.py
```diff
@@ -1,18 +1,116 @@
...
+    """
+    log_spec_str = log_spec_str.lower()
```
tomskikh: Can target only be in lowercase?

savant/utils/logging/log_utils.py
tomskikh: Fix this in other places too

tomskikh (changes requested):

savant/utils/logging/log_utils.py
abramov-oleg: Fixed.

savant/utils/logging/log_utils.py
abramov-oleg: You're right, it's possible that log targets will differ only in case, even though it's unconventional.

Fixed.

savant/utils/logging/log_utils.py
```diff
@@ -1,18 +1,116 @@
...
+        if log_directive:
+
+            eq_num = log_directive.count('=')
+
+            if eq_num == 1:
+                target, level = log_directive.split('=')
+            elif eq_num == 0:
+                log_directive = log_directive.lower()
+                # no = sign, i.e. only target or only level
+                if log_directive in LOG_LEVEL_STR_TO_RS:
```
tomskikh:
```suggestion
                log_directive = log_directive
                # no = sign, i.e. only target or only level
                if log_directive.lower() in LOG_LEVEL_STR_TO_RS:
```

abramov-oleg: Removed buggy assignment line.

tomskikh (approved):

PR #439, opened 2023-09-21
Prevent object parent recursion

denisvmedyantsev:


savant/deepstream/buffer_processor.py
```diff
@@ -606,7 +606,9 @@ def prepare_element_output(self, element: PipelineElement, buffer: Gst.Buffer):
                         # for object/complex models output - `bbox_tensor` and
                         # `selected_bboxes` - indices of selected bboxes and meta
                         # for attribute/complex models output - `values`
-                        bbox_tensor, selected_bboxes, values = None, None, None
+                        bbox_tensor: Optional[np.array] = None
+                        selected_bboxes: Optional[list] = None
```
dorgun:
```suggestion
                        selected_bboxes: Optional[List] = None
```
Since we're still using python 3.8, it's better to use typing generics

savant/deepstream/buffer_processor.py
```diff
@@ -606,7 +606,9 @@ def prepare_element_output(self, element: PipelineElement, buffer: Gst.Buffer):
                         # for object/complex models output - `bbox_tensor` and
                         # `selected_bboxes` - indices of selected bboxes and meta
                         # for attribute/complex models output - `values`
-                        bbox_tensor, selected_bboxes, values = None, None, None
+                        bbox_tensor: Optional[np.array] = None
```
dorgun:
```suggestion
                        bbox_tensor: Optional[np.ndarray] = None
```

savant/deepstream/buffer_processor.py
```diff
@@ -606,7 +606,9 @@ def prepare_element_output(self, element: PipelineElement, buffer: Gst.Buffer):
...
+                        selected_bboxes: Optional[list] = None
+                        values: Optional[list] = None
```
dorgun:
```suggestion
                        values: Optional[List] = None
```

savant/deepstream/meta/object.py
```diff
@@ -75,20 +75,11 @@ def __init__(
...
-                parent.object_meta_impl.ds_object_meta if parent is not None else None,
-            )
-        else:
-            raise MetaValueError(
-                f'parent is `{type(parent)}` type, expected `{self.__class__.__name__}`'
-            )
+        pyds.nvds_add_obj_meta_to_frame(
+            self._frame_meta.frame_meta,
+            self.ds_object_meta,
+            parent.object_meta_impl.ds_object_meta if parent is not None else None,
```
dorgun, line 81: Why did you remove the check? The user can simply create an ObjectMeta and specify it as a parent. When an ObjectMeta is created, it is not immediately added to the DS meta and therefore its object_meta_impl will be None. Eventually object_meta_impl will be None and there will be an error.

savant/deepstream/meta/object.py
```diff
@@ -295,21 +289,25 @@ def element_name(self) -> str:
...
         :param frame_meta: Deepstream frame meta.
+        :param depth: Object parent recursion depth.
         :return:
         """
         self = cls.__new__(cls)
         self.ds_object_meta = object_meta
         self._frame_meta = frame_meta
         self._bbox = None
-        if not nvds_is_empty_object_meta(object_meta.parent):
+        if not nvds_is_empty_object_meta(object_meta.parent) and depth > 0:
```
dorgun, line 308: My suggestion is to just move this code to getter parent and make it lazy to create a wrapper for parent. Then we won't need to check the recursion depth. Or at least give some warning here. But now it is just none when it is deeper than 2 and that's all. What if a user actually creates a relation with a depth greater than 2?

dorgun (changes requested):

savant/deepstream/meta/object.py
bwsw: @denisvmedyantsev @dorgun ignore that, yet. Not worth doing extra engineering.

savant/deepstream/meta/object.py
denisvmedyantsev: Because a line above `self.parent = parent` (parent setter) checks the same.

denisvmedyantsev: Because a line above `self.parent = parent` (parent setter) checks the same.

savant/deepstream/buffer_processor.py
denisvmedyantsev: done

savant/deepstream/buffer_processor.py
denisvmedyantsev: done

savant/deepstream/buffer_processor.py
denisvmedyantsev: done

dorgun (approved):

PR #483, opened 2023-10-09
Face Reid sample Index builder uses Client SDK

### LINKED ISSUE #449: Change Index builder from Adapter to Client SDK in face_reid sample
Enable sending shutdown command on complete.

abramov-oleg:


samples/face_reid/README.md
```diff
@@ -14,7 +14,7 @@ The sample is split into two parts: Index Builder and Demo modules.
 
 Index builder module loads images from [gallery](./assets/gallery), detects faces and facial landmarks, performs face preprocessing and facial recognition model inference. The resulting feature vectors are added into [hnswlib](https://github.com/nmslib/hnswlib) index, and the index (along with cropped face images from gallery) is saved on disk in the `index_files` directory.
 
-Note, when adding new gallery images it is important to make sure that they are as close to 16:9 aspect ratio as possible. The reason being that Index Builder pipeline processes all images in a single resolution with 16:9 aspect ratio, and resizing may introduce image warping that will negatively affect both detector and ReID models' performance.
+Note that the Index Builder pipeline processes all images in a single resolution with 16:9 aspect ratio, but the gallery is sent to the pipeline through the Client SDK which can take care of padding the image to a target aspect ratio. As such it's safe to add new images of any aspect ratio to the gallery. The gallery images are still expected to contain 1 face each and to be named according to scheme `<person_name>_<img_n>.jpeg`.
```
denisvmedyantsev: > The gallery images are still expected..

I'm not sure the word "still" is necessary here.

samples/face_reid/src/index_builder_client.py
```diff
@@ -0,0 +1,162 @@
...
+        for img_path in pathlib.Path(args.gallery_dir).glob('*.jpeg')
+    ]
+
+    source_ids = set()
+    for src_jpeg in src_jpegs:
+        source(src_jpeg, send_eos=False)
+        source_ids.add(src_jpeg.source_id)
+
+    for src_id in source_ids:
+        source.send_eos(src_id)
```
denisvmedyantsev: Why do we send eos in a separate cycle?

savant/client/frame_source/jpeg.py
```diff
@@ -87,18 +112,61 @@ def with_framerate(self, framerate: Tuple[int, int]) -> 'FrameSource':
...
+                right = self._width - width - left
+            else:
+                left = right = 0
+
+            img = cv2.copyMakeBorder(
+                img, top, bottom, left, right, cv2.BORDER_CONSTANT, value=[0] * channels
+            )
+        else:
+            # resize
+            img = cv2.resize(img, (self._width, self._height))
```
denisvmedyantsev: Shouldn't we be concerned about the aspect ratio here? (Letterbox)

samples/face_reid/README.md
denisvmedyantsev: Wouldn't it be more convenient and familiar to create a folder person_name and put the images (any_name.jpg) in it?

samples/face_reid/README.md
```diff
@@ -48,24 +48,18 @@ docker compose -f docker-compose.x86.yml --profile index up
...
-INFO ... > The pipeline is initialized and ready to process data...
+INFO ... > The pipeline is initialized and ready to process data. Initialization took ...
+INFO ... > Setting module status to ModuleStatus.RUNNING
 ```
 
-After that the `pictures-source` container will be started automatically.
+After that the `index-builder-client` container will be started automatically.
 
-Index Builder module does not stop automatically and requires manual exit. It is safe to do so once the `face_reid-pictures-source` container exits and `face_reid-index-builder` container logs the message following messages (assuming the default gallery image set)
+The `index-builder-client` service runs the [index_builder_client.py](./src/index_builder_client.py) script, which loads the images from the [gallery](./assets/gallery), pads them to 16:9 aspect ratio if necessary, sends them to the `index-builder-pipeline` and uses the received results to create `index_files/index.bin` index file and the cropped face images in the `index_files/processed_gallery` dir.
```
denisvmedyantsev: What is the reason for using a 16:9 aspect ratio (portrait usually has a different aspect ratio)? 

You can use the detector's aspect ratio and resolution.

samples/face_reid/src/index_builder_client.py
```diff
@@ -0,0 +1,162 @@
...
+import hnswlib
+import numpy as np
+
+from samples.face_reid.utils import pack_person_id_img_n
+from savant.client import JpegSource, SinkBuilder, SourceBuilder
+from savant.utils.logging import get_logger, init_logging
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
```
denisvmedyantsev, line 22: Do we really need all these arguments as script parameters? Why not set them as constants?

savant/client/frame_source/jpeg.py
```diff
@@ -26,6 +27,12 @@ class JpegSource(FrameSource):
 
     :param source_id: Source ID.
     :param filepath: Path to a JPEG file.
+    :param pts: Frame presentation timestamp.
+    :param framerate: Framerate (numerator, denominator).
+    :param source_id_add_size_suffix: Whether to add a suffix
```
denisvmedyantsev: This parameter (source_id_add_size_suffix) looks strange. Could you explain the reason for it?

denisvmedyantsev (changes requested):
Let's discuss PR at the meeting. I can't understand some of the solutions you have implemented.

samples/face_reid/README.md
abramov-oleg: Removed "still".

Regarding the gallery dir structure, personally, I don't think it matters. Should I reorganize it?

samples/face_reid/src/index_builder_client.py
abramov-oleg: This is done in order to send EOS only once for each source_id.

The only way to avoid SeqId discrepancy warnings from `savant_rs::message::validate_seq_iq` (both from the pipeline and the client) that I've found is to break up the images into several "streams", i.e. source_ids, in each of which all the frames have the same size, and to send EOS only once for each such "stream".

bwsw: @abramov-oleg does EOS after each image leads to the warning message?

ah, I see what happens. Will think how to deal with it :)

savant/client/frame_source/jpeg.py
abramov-oleg: Added resizing while preserving aspect ratio.

savant/client/frame_source/jpeg.py
abramov-oleg: Removed it.

samples/face_reid/README.md
abramov-oleg: Changed to use the detector's resolution. Updated README.

bwsw:
@abramov-oleg when EOS is sent, use the `clear_source_seq_id` for the source to reset its counter:

https://insight-platform.github.io/savant-rs/modules/savant_rs/utils_serialization.html#savant_rs.utils.serialization.clear_source_seq_id

It must fix WARNs if sending EOS after every image. However, sending EOS after each image may be inefficient as it causes excessive Gstreamer operations. Mabe to keep the functional model using a single source, but adding an attribute with "with_update" including the relevant person information.

Available in `0.1.71`.

savant/client/frame_source/jpeg.py
```diff
@@ -88,11 +93,14 @@ def with_update(self, update: VideoFrameUpdate) -> 'JpegSource':
...
 
     def build_frame(self) -> Tuple[VideoFrame, bytes]:
-        # TODO: get image size without decoding
-        img = cv2.imread(self._filepath)
-        height, width, _ = img.shape
-        with open(self._filepath, 'rb') as f:
-            content = f.read()
+        width, height = get_jpeg_size(self._file)
+
+        if hasattr(self._file, 'read'):
```
denisvmedyantsev: I think it is better to find out the type of `file` argument in __init__, set some internal attribute, and do not mix different ways of checking if it is a file path or handler in the code later.

denisvmedyantsev (approved):

savant/client/frame_source/jpeg.py
abramov-oleg: That introduces mypy typing errors and the possibility of the flag being out of sync with the `_file`  attribute's contents.

Changed the check for the type of value in `file` to be consistent.

PR #486, opened 2023-10-11
Fix pipeline hanging on error

### LINKED ISSUE #466: Face ReID demo hangs on error
It can be reproduced under the conditions described in #465 or under the conditions described below.

Use, for example, `./scripts/run_module.py` to start the demo with `module_performance.yml` config. 

To reproduce, modify the face_preproc.py, and raise an exception. If an exception occurs, the module will not terminate after the source has been processed and the module container will go to the "unhealthy" status after some time.

tomskikh:


savant/entrypoint/main.py
```diff
@@ -89,3 +89,7 @@ def main(config_file_path: str):
                 os._exit(1)  # pylint: disable=protected-access
     except Exception as exc:  # pylint: disable=broad-except
         logger.error(exc, exc_info=True)
+        exit(1)
```
denisvmedyantsev, line 92: What is the reason for `exit` here? `main` is the only function in entrypoint and there it's your final row

savant/entrypoint/main.py
```diff
@@ -89,3 +89,7 @@ def main(config_file_path: str):
...
+
+    if runner.error is not None:
```
denisvmedyantsev, line 94: `runner` is out of its context?

denisvmedyantsev (changes requested):

savant/entrypoint/main.py
tomskikh: `exit` here is needed to stop the module with non-zero exit code.

savant/entrypoint/main.py
tomskikh: We don't need context here.

denisvmedyantsev (approved):

PR #493, opened 2023-10-12
Updated cmake version and refactoring cmake files in savantboost

### LINKED ISSUE #399: Refactor SavantBoost
1. Refactor the code, reduce warnings
2. Refactor CMakeList, eg. https://cmake.org/cmake/help/latest/module/FindCUDA.html
3. Check options for build for CUDA 12+

dorgun:


docker/Dockerfile.deepstream
```diff
@@ -45,7 +45,7 @@ RUN python setup.py bdist_wheel && rm -rf _skbuild
 FROM base_builder AS savantboost_build
 
 # newer cmake ver. needed for "FindCUDAToolkit"
```
denisvmedyantsev: Do we still need `newer cmake ver. for "FindCUDAToolkit"`?

denisvmedyantsev:
Did you leave hundreds of `gnu_printf` warnings for some reason?
```
#26 4.580 /usr/include/glib-2.0/glib/gerror.h(54): warning #1217-D: unrecognized format function type "gnu_printf" ignored
#26 4.580                                   ...) __attribute__((__format__ (gnu_printf, 3, 4)));
#26 4.580                                                                   ^
#26 4.580 
#26 4.581 Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
#26 4.581 
```

denisvmedyantsev:
We can add ninja in `base_builder` to avoid the first step fails for savantboost and pygstsavantframemeta `Trying 'Ninja' generator - failure`. Up to you.

docker/Dockerfile.deepstream
```diff
@@ -45,7 +45,7 @@ RUN python setup.py bdist_wheel && rm -rf _skbuild
 FROM base_builder AS savantboost_build
```
denisvmedyantsev, line 44: There are 2 TODOs above. 

The first one (`TODO: Update library code, CMakeLists (see build process warnings)`) will not be relevant after your refactoring. 

The second `TODO: Check Jetson requirements (Xavier t194, Orin t234)` is still actual. What do you think about installing nvidia-l4t-multimedia from the t194 repo, would there be any problems using savantboost on Orin (t234)? What components do we really need to build savantboost on Jetson?

docker/Dockerfile.deepstream
dorgun: Deleted this comment

docker/Dockerfile.deepstream
dorgun: Without the nvidia-l4t-multimedia library, you can't build a library. On Orin I checked it builds fine, age-gender example using custom preprocessing works fine.

dorgun:
The problem with the gnu_printf warning has been solved.
Regarding Ninja. scikit-build used to build the python package, just check it first, then Unix Makefiles. I don't see the point of installing Ninja just to remove this message from the logs.

docker/Dockerfile.deepstream
denisvmedyantsev: If there is no problem with Orin, remove the second TODO too.

denisvmedyantsev (approved):

PR #501, opened 2023-10-17
Video pass-through processing

### LINKED ISSUE #443: Video pass-through processing
When video pass-through is configured, no encoding happens: the video is just copied through the pipeline (it can be relayed into the frame as `internal` and, at the end of the pipeline, placed back as external to ZMQ). 

```
parameters:
  output_frame:
    codec: copy
```

When the video pass-through mode is configured, `draw_func` doesn't have an effect. We can allow the user to use it, but we need to **WARN** about the restriction. Warning may look like this: 

`The pipeline is configured in video pass-through mode. In this mode, frame modifications exist only in the pipeline but are not propagated through the sinks.`

In this mode, conditional video encoding doesn't have an effect. If it is configured, we must result in a **WARN** message:

`Conditional video encoding is configured for the pipeline, which is incompatible with pass-through mode and will be ignored.`

The geometry for objects must be re-calculated to map incoming frame properties.

The in-pipeline transformations must be wiped from the frame meta.

### Sample

The sample demonstrates a chained pipeline based on one re-id pipeline:

Each step represents a separate docker container (use element group to place all of them in as single yaml):

```
loop > face-detector > reid > draw > ao-rtsp 
```

Pass-through containers:
- face-detector
- re-id

tomskikh:


samples/pass_through_processing/docker-compose.x86.yml
```diff
@@ -0,0 +1,125 @@
...
+    image: ghcr.io/insight-platform/savant-deepstream:latest
+    restart: unless-stopped
+    volumes:
+      - zmq_sockets:/tmp/zmq-sockets
+      - ../../models/peoplenet_detector:/models
+      - ../../downloads/peoplenet_detector:/downloads
+      - ..:/opt/savant/samples
+    command: samples/pass_through_processing/module.yml
+    environment:
+      - MODULE_NAME=pass-through-processing-detector
```
denisvmedyantsev: Using only one name parameter for distinct modules, MODULE_STAGE or MODULE_NAME, is enough.

samples/pass_through_processing/docker-compose.x86.yml
```diff
@@ -0,0 +1,125 @@
...
+    volumes:
+      - zmq_sockets:/tmp/zmq-sockets
+      - ../../models/peoplenet_detector:/models
+      - ../../downloads/peoplenet_detector:/downloads
+      - ..:/opt/savant/samples
+    command: samples/pass_through_processing/module.yml
+    environment:
+      - MODULE_NAME=pass-through-processing-tracker
+      - MODULE_STAGE=tracker
+      - BATCH_SIZE=1
```
denisvmedyantsev: Set `parameteres.batch_size: 1` in module config. There is no need to pass it via env.

savant/deepstream/pipeline.py
```diff
@@ -85,6 +86,8 @@ def __init__(
         pipeline_cfg: Pipeline,
         **kwargs,
     ):
+        self._logger = get_logger(name)
```
denisvmedyantsev: Logger initialization is already done in parent GstPipeline.

samples/pass_through_processing/docker-compose.x86.yml
tomskikh: Using only MODULE_STAGE

samples/pass_through_processing/docker-compose.x86.yml
tomskikh: Fixed

savant/deepstream/pipeline.py
tomskikh: I've needed the logger before initialization in GstPipeine to log a warning. Moved a warning in module_config.py.

docs/source/savant_101/12_video_processing.rst
```diff
@@ -492,3 +493,23 @@ To configure conditional encoding, add a subsection to ``output_frame`` as follo
...
+To configure pass-through mode, set ``output_frame.codec`` to ``copy``:
+
+.. code-block:: yaml
+
+    output_frame:
+      codec: copy
+
+.. note::
+
+    Conditional encoding is ignored in pass-through mode.
```
bwsw: @tomskikh please revert that. I was mistaken. Let us keep conditional video encoding for pass-through mode too.

bwsw (changes requested):

savant/deepstream/pipeline.py
```diff
@@ -108,6 +108,7 @@ def __init__(
         init_telemetry(name, telemetry)
 
         output_frame = kwargs.get('output_frame')
+        self._pass_through_mode = bool(output_frame) and output_frame['codec'] == 'copy'
```
denisvmedyantsev, line 111: unnecessary bool()

gst_plugins/python/savant_rs_video_demux.py
```diff
@@ -424,11 +414,91 @@ def handle_video_frame(
...
+            video_frame.source_id,
+            frame_pts,
         )
         return result
 
+    def _build_frame_buffer(
+        self,
+        video_frame: VideoFrame,
+        buffer: Gst.Buffer,
+    ) -> Union[Gst.Buffer, Gst.FlowReturn]:
```
denisvmedyantsev: May be better to check content.is_none() out of function and to return only buffer or raise an exception, and wrap function call
```
try:
    frame_buf = self._build_frame_buffer(video_frame, buffer)
    ...
    return Gst.FlowReturn.OK
except ..:
    ...
    return Gst.FlowReturn.ERROR
```

denisvmedyantsev (approved):
Don't forget to check perf.

gst_plugins/python/savant_rs_video_demux.py
tomskikh: Wrapped in try/except.

savant/deepstream/pipeline.py
tomskikh: We need it to be strictly bool since the value is set as a property of `zeromq_source_bin`. Otherwise the value could be `{}`.

docs/source/savant_101/12_video_processing.rst
tomskikh: Reverted

bwsw (approved):

Issue #495, opened 2023-10-13
Module processes only one source when multi-stream adapter is restarted


denisvmedyantsev:
Module config
```
name: test
parameters:
  batch_size: 1
```

The first run of multi-stream adapter with 8 sources:
```
./scripts/run_source.py multi-stream --number-of-streams=8 data/jumanji_cast.mp4
```
Adapter sends 3725 frames x8, Module processes 8 sources, 29800 frames.

Module log fragment
```
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-0
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-0 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-0: src_source-0.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-1
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-1 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-1: src_source-1.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-2
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-2 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-2: src_source-2.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-3
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-3 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-3: src_source-3.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-4
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-4 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-4: src_source-4.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-5
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-5 added
 INFO  insight::savant::savant_rs_video_demux       > Created new src pad for source source-5: src_source-5.
 INFO  insight::savant::savant_rs_video_decode_bin  > Adding branch with source source-6
 INFO  insight::savant::savant_rs_video_decode_bin  > Branch with source source-6 added
[...]

PR #506, opened 2023-10-18
Fix module processing of restarted multi-stream adapter

tomskikh:


savant/utils/zeromq.py
```diff
@@ -359,7 +359,7 @@ def filter(self, routing_id: Optional[bytes], topic: bytes):
 
         if topic not in self.routing_ids:
             self.routing_ids[topic] = routing_id
-            self.routing_ids_cache[routing_id] = None
+            self.routing_ids_cache[(topic, routing_id)] = None
 
         elif self.routing_ids[topic] != routing_id:
             if routing_id in self.routing_ids_cache:
```
denisvmedyantsev: This will be false every time.

denisvmedyantsev (changes requested):

savant/utils/zeromq.py
tomskikh: Fixed

denisvmedyantsev (approved):

bwsw:
Closes #495

PR #520, opened 2023-10-24
507 add tracker postprocessing function

### LINKED ISSUE #507: Add tracker postprocessing function
The function, by default, removes objects which the tracker creates.

abramov-oleg:


libs/gstsavantframemeta/gstsavantframemeta/include/savantnvprobes.h
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * A probe for the nvtracker src pad to remove the objects created by the nvtracker element.
+ *
+ * @param pad GStreamer pad.
+ * @param info Pad probe info.
+ * @param user_data Pointer to user data.
+ */
+GstPadProbeReturn remove_tracker_objs_pad_probe(GstPad *pad,
+                                             GstPadProbeInfo *info,
+                                             gpointer user_data);
```
tomskikh: Reformat the code, the args are not aligned.

libs/gstsavantframemeta/gstsavantframemeta/src/savantnvprobes.cpp
```diff
@@ -0,0 +1,55 @@
...
+            pad,
+            "Tracker obj remover: skipping buffer %ld, batch is empty.",
+            buffer->pts);
+        return GST_PAD_PROBE_PASS;
+    }
+
+    NvDsFrameMetaList *l_frame = NULL;
+    for (l_frame = batch_meta->frame_meta_list; l_frame != NULL; l_frame = l_frame->next) {
+        NvDsFrameMeta *frame_meta = (NvDsFrameMeta *) (l_frame->data);
+        std::vector<NvDsObjectMeta*> removal_list;
```
tomskikh, line 27: Looks like it would cause a memory leak. Is this list cleared automatically?

libs/gstsavantframemeta/gstsavantframemeta/src/savantnvprobes.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+#include "gstnvdsmeta.h"
+
+
+GstPadProbeReturn remove_tracker_objs_pad_probe(GstPad *pad,
+                                                GstPadProbeInfo *info,
+                                                gpointer user_data) {
+    GstBuffer *buffer = GST_PAD_PROBE_INFO_BUFFER(info);
+    if (!buffer) {
+        GST_INFO_OBJECT(pad, "Tracker obj remover: skipping NULL buffer.");
```
tomskikh: The log entry contains filename, line and name of method. There's no need to add a prefix to the log lines.

libs/gstsavantframemeta/pygstsavantframemeta/__init__.py
```diff
@@ -173,3 +173,11 @@ def add_pad_probe_to_unpack_and_move_batch(
         video_pipeline.memory_handle,
         stage,
     )
+
+
+def add_pad_probe_to_remove_tracker_objs(element: Gst.Element):
```
tomskikh: Why this method accepts `Gst.Element` and not `Gst.Pad` like the other `add_pad_probe_` methods?

savant/gstreamer/element_factory.py
```diff
@@ -92,3 +97,20 @@ def on_pad_removed(elem: Gst.Element, pad: Gst.Pad):
         src_element.link(decodebin)
 
         return src_decodebin
+
+    @staticmethod
+    def create_nvtracker(element: PipelineElement) -> Gst.Element:
```
tomskikh: Move this to `NvDsElementFactory`.

tomskikh (changes requested):
Document the feature.

libs/gstsavantframemeta/gstsavantframemeta/src/savantnvprobes.cpp
abramov-oleg: The vector is allocated on the stack and does not contain any pointers to new allocations. What's at risk of being leaked?

I assume that NvDsObjectMetas are properly released in the `nvds_remove_obj_meta_from_frame` call.

libs/gstsavantframemeta/gstsavantframemeta/include/savantnvprobes.h
abramov-oleg: Reformatted.

libs/gstsavantframemeta/gstsavantframemeta/src/savantnvprobes.cpp
abramov-oleg: Removed the logging prefix.

savant/gstreamer/element_factory.py
abramov-oleg: Moved to DS factory class.

libs/gstsavantframemeta/pygstsavantframemeta/__init__.py
abramov-oleg: I see no reason leaving `gst_element_get_static_pad` call on the python side.

abramov-oleg: Is the comment about the naming?

libs/gstsavantframemeta/gstsavantframemeta/src/savantnvprobes.cpp
tomskikh: > The vector is allocated on the stack and does not contain any pointers to new allocations.

In this case that's ok.

libs/gstsavantframemeta/pygstsavantframemeta/__init__.py
tomskikh: > I see no reason leaving `gst_element_get_static_pad` call on the python side.

* The method `add_pad_probe_to_remove_tracker_objs` basically a wrapper for `gst_pad_add_probe`, so it makes sense to pass only pad, not the whole element.
* It would be consistent with the other methods to pass only a pad. For any method to add a pad probe we are calling `gst_element_get_static_pad` on the python side.

> Is the comment about the naming?

No, the name is ok.

abramov-oleg: Refactored the bindings to pass a pointer to pad from the python side.

tomskikh (approved):

PR #525, opened 2023-10-25
Deduplicate video data when video is pass-through in Kafka-Redis sink adapter

### LINKED ISSUE #518: Deduplicate video data when video is pass-through in Kafka-Redis sink adapter
When Kafka-Redis sink is configured with Redis and video is processed in the pass-through mode, the only need is to fix TTL, but the same reference (use UUID) can be used.

The deduplication works only if the data location remains intact. TTL is updated as necessary.

The implementation can be as follows:

The user configures `DEDUPLICATE` variable specifying whether deduplication has a place or not. When `DEDUPLICATE` is `true` and the video is copied, the sink first tries to update the video TTL with `EXPIRE`. If it fails (returns 0, according to Redis spec) then it uploads the video with TTL.

tomskikh:


adapters/python/sinks/kafka_redis.py
```diff
@@ -215,30 +224,81 @@ async def store_frame_content(
...
+
+    async def need_to_store_content_to_redis(
+        self,
+        frame: VideoFrame,
+        content_key: str,
+        location: str,
+    ):
+        """Check if the frame content needs to be stored to Redis.
+
+        The frame content doesn't need tobe stored in Redis when the following conditions are met:
```
denisvmedyantsev: "The frame content doesn't need tobe..." => "The frame content doesn't need to be..."

denisvmedyantsev (approved):

PR #536, opened 2023-10-30
317 super resolution demo

### LINKED ISSUE #317: Image upscale/sr demo
Convert 640x480 to FHD

denisvmedyantsev:


samples/super_resolution/README.md
```diff
@@ -0,0 +1,29 @@
+# Super Resolution
+
+The pipeline shows the use of super-resolution neural networks to improve the quality and resolution of the input video. 
+
+We prepared ONNX versions (using standard `torch.onnx.export`) of two lightweight NinaSR models ninasr_b0 and ninasr_b1 with different scales (x2/3/4) from the [TorchSR repository](https://github.com/Coloquinte/torchSR). The demo uses input video in 360p format and model ninasr_b0 with scale 3 to output video in 1080p format. The model and scale can be changed in the module configuration [file](module.yml).
+
+The demo is prepared for use on dGPU. Separately, the module can be run on the Jetson Xavier platform also.
```
abramov-oleg: Don't understand use of `separately` here.

samples/super_resolution/docker-compose.x86.yml
```diff
@@ -0,0 +1,69 @@
...
+      - "888:888"    # HLS
+      - "8889:8889"  # WebRTC
+    volumes:
+      - zmq_sockets:/tmp/zmq-sockets
+      - ../assets/stub_imgs:/stub_imgs
+    environment:
+      - ZMQ_ENDPOINT=sub+connect:ipc:///tmp/zmq-sockets/output-video.ipc
+      - SOURCE_ID=video
+      - FRAMERATE=25/1
+      - STUB_FILE_LOCATION=/stub_imgs/smpte100_1280x720.jpeg
```
abramov-oleg: Maybe use a larger stub? Right now it is difficult to evaluate the result, since it is so severely downsized.

samples/super_resolution/overlay.py
```diff
@@ -0,0 +1,74 @@
...
+            # transform super resolution and place on the frame
+            if sr_attr:
+                sr_image_np = sr_attr.value.clip(0.0, 1.0)
+                sr_image_np = (sr_image_np * 255).astype(np.uint8)
+                # chw => hwc
+                sr_image_np = np.transpose(sr_image_np, (1, 2, 0))
+                # rgb => rgba
+                sr_image_np = np.dstack(
+                    (sr_image_np, np.zeros(SUPER_RESOLUTION[::-1], dtype=np.uint8))
+                )
```
abramov-oleg: This sets the image alpha to 0, which is supposed to make it fully transparent. We avoid the bug only because the conversion to GpuMat below resets the alpha to 255, for no obvious reason.

Suggest to use opencv.cvtColor.

samples/super_resolution/overlay.py
```diff
@@ -0,0 +1,74 @@
...
+                sr_image = cv2.cuda_GpuMat(sr_image_np)
```
abramov-oleg: This implies a blocking upload of image to gpu. May not be very important, but since we try to use async ops wherever possible, I suggest to break the operation up into 2 steps: allocate a gpumat, then async upload (look for example in the `Artist.__exit__`)

Additionally, it may be worth it to modify `Artist.add_graphic()` to accept numpy array, i.e. hide the conversion from the user inside the method.

abramov-oleg (changes requested):

samples/super_resolution/README.md
denisvmedyantsev: Separate = without other performance-impacting demo services (e.g. AO-RTSP). But I'd better delete that confusing sentence.

samples/super_resolution/docker-compose.x86.yml
denisvmedyantsev: replaced with smpte100_3840x1080.jpeg

samples/super_resolution/overlay.py
denisvmedyantsev: alpha 0 is my fault, replaced with 255. Why is it better to use opencv.cvtColor?

samples/super_resolution/overlay.py
denisvmedyantsev: reasonable comment, fixed in artist

abramov-oleg (approved):

samples/super_resolution/overlay.py
abramov-oleg: More readable, less bug prone.

Up to you.

PR #539, opened 2023-11-01
Implement GPU-less Always-On RTSP sink adapter

### LINKED ISSUE #527: GPU-less AO-RTSP
If Nvidia runtime is not detected, the AO-RTSP sink must use CPU-based decoders/encoders.
Switch all `l4t` demos to use GPU-less AO-RTSP with RAW-RGBA output (modify it through env var).

tomskikh:


adapters/ds/sinks/always_on_rtsp/utils.py
```diff
@@ -0,0 +1,15 @@
...
+import cv2
+import numpy as np
+
+from savant.gstreamer import Gst
+
+Frame = Union[cv2.cuda.GpuMat, np.ndarray, bytes]
+"""Type alias for a frame content."""
+
+
+def nvidia_runtime_is_available() -> bool:
```
denisvmedyantsev, line 14: I think it's better to cache the call here.

adapters/ds/sinks/always_on_rtsp/timestamp_overlay.py
```diff
@@ -25,18 +26,27 @@ def __init__(self):
...
             'Placing timestamp %s on a frame of a size %sx%s',
             timestamp,
-            frame_width,
-            frame_height,
+            width,
+            height,
         )
-        self._placeholder.fill(0)
+        if isinstance(frame, cv2.cuda.GpuMat):
+            placeholder = self._placeholder
```
denisvmedyantsev, line 44: It's better, in my opinion, to move here from init `np.zeros((self._height, self._width, 4), dtype=np.uint8)` and to remove `self._placeholder`.

adapters/ds/sinks/always_on_rtsp/config.py
```diff
@@ -78,6 +79,34 @@ def __init__(self):
             'in format WIDTHxHEIGHT, for example 1920x1080.'
         )
 
+        self._nvidia_runtime_is_available: Optional[bool] = None
```
denisvmedyantsev: Why not set fields directly in init and avoid checking None in property calls?

adapters/ds/gst_plugins/python/always_on_rtsp_frame_processor.py
```diff
@@ -182,26 +280,57 @@ def crop_to_fit(
...
+
+        return Gst.FlowReturn.OK
+
+    @contextmanager
+    def _gst_buffer_to_np_array(self, buffer: Gst.Buffer):
+        """Convert Gst.Buffer to numpy array.
+        All the changes will be applied to the buffer.
+        """
+
+        is_mapped, map_info = buffer.map(Gst.MapFlags.READ | Gst.MapFlags.WRITE)
```
denisvmedyantsev: There are ctypes ffi and `map_gst_buffer` in `savant/gstreamer`. Are they still relevant? - Then please use (modify and use) them, otherwise delete them.

denisvmedyantsev (commented):
I leave some comments without `approve`. Will approve after I test the samples.

denisvmedyantsev (approved):

adapters/ds/sinks/always_on_rtsp/timestamp_overlay.py
tomskikh: We don't need to create `self._placeholder` every time in this case. We can create it only once and reuse it.

adapters/ds/sinks/always_on_rtsp/config.py
tomskikh: `nvidia_runtime_is_available()` won't work here since `Gst.Init()` is called after the init.

adapters/ds/sinks/always_on_rtsp/utils.py
tomskikh: Made this method cached.

adapters/ds/sinks/always_on_rtsp/config.py
tomskikh: Removed `nvidia_runtime_is_available` property from config and made `converter` and `video_raw_caps` properties cached.

adapters/ds/gst_plugins/python/always_on_rtsp_frame_processor.py
tomskikh: Used `map_gst_buffer`. Now we don't need to copy memory to make a writable numpy buffer.

PR #545, opened 2023-11-06
Fix healthcheck script path.

### LINKED ISSUE #541: Template sample native docker healthcheck doesn't work
How to reproduce:

```
docker compose -f samples/template/docker-compose.x86.yml up jaeger -d
docker compose -f samples/template/docker-compose.x86.yml up module -d
docker ps -a
```

`template-module-1` container is not advancing from health status `starting` to `healthy`.

The problem is that the `samples/template/docker/Dockerfile.*` change WORKDIR which makes the HEALTHCHECK script path from base image to be incorrect.

Either don't change WORKDIR or set absolute path for HEALTHCHECK in the base Dockerfile.

denisvmedyantsev:


samples/template/docker-compose.x86.yml
```diff
@@ -29,6 +29,13 @@ services:
       - --socket
       - pub+connect:ipc:///tmp/zmq-sockets/input-video.ipc
       - --sync
+    deploy:
```
bwsw, line 32: Why this resource is required for the module input?
It looks to me, that the declaration in the `runtime: nvidia` is excessive in the L4T image because the URI input does not use the Nvidia runtime, despite it exists in the DS-container.

bwsw (changes requested):

samples/template/docker-compose.x86.yml
denisvmedyantsev: I added it to avoid import errors from some plugins, e.g.
```
template-uri-input-1       | Traceback (most recent call last):
template-uri-input-1       |   File "/opt/savant/gst_plugins/python/frame_tag_filter.py", line 5, in <module>
template-uri-input-1       |     import pyds
template-uri-input-1       | ImportError: libcuda.so.1: cannot open shared object file: No such file or directory
template-uri-input-1       | 
template-uri-input-1       | (gst-plugin-scanner:10): GStreamer-WARNING **: 08:43:55.977: Failed to load plugin '/usr/lib/x86_64-linux-gnu/gstreamer-1.0/deepstream/libnvdsgst_infer.so': libcuda.so.1: cannot open shared object file: No such file or directory
```

bwsw: @denisvmedyantsev OK

bwsw (approved):

bwsw:
Approved.

PR #552, opened 2023-11-08
Add build model engine entrypoint

### LINKED ISSUE #336: Improve how the models are compiled in isolated pipelines run before the main
Now the process of model building and validation is highly obfuscated. We need a more transparent approach.

denisvmedyantsev:


samples/template/src/module/run.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 import os
 
-from savant.entrypoint.main import main
+from savant.entrypoint.main import run_module
 
 if __name__ == '__main__':
-    main(os.path.join(os.path.dirname(__file__), 'module.yml'))
+    run_module(os.path.join(os.path.dirname(__file__), 'module.yml'))
```
bwsw: 1. The idea is not good. It requires changing videos, docs, manuals, screenshots. I disapprove. Change it.
2. Docs missing showing how to compile models without running the processing.

bwsw (changes requested):

samples/template/src/module/run.py
denisvmedyantsev: 1) Rolled back the renaming of main.

denisvmedyantsev: 2. Added.

bwsw (approved):

PR #563, opened 2023-11-20
Refactor nvinfer processing

### LINKED ISSUE #550: Savant CuPy integration
Demonstrate Savant CuPY integration on the base of the YOLOV8-Seg demo.

denisvmedyantsev:
#550

### DIFF
--- a/savant/deepstream/nvinfer/processor.py
+++ b/savant/deepstream/nvinfer/processor.py
@@ -84,7 +84,7 @@ class NvInferProcessor:
                 self._frame_rect[3] + frame_params.padding.top,
             )
 
-        # video pipeline (frame/source info, span etc.)
+        # video pipeline (frame/source info, telemetry span etc.)
         self._video_pipeline = video_pipeline
 
         self._element_name = element.name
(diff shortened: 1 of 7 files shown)


savant/deepstream/nvinfer/processor.py
```diff
@@ -0,0 +1,614 @@
...
+                            element_name=self._element_name,
+                            name=attr.name,
+                            value=label_info.result_label,
+                            confidence=label_info.result_prob,
+                        )
+        self._restore_frame(buffer)
+
+    def _logger_trace(self, msg: str, *args, **kwargs):
+        if self._logger.isEnabledFor(logging.TRACE):
+            self._logger.trace(msg, args, **kwargs)
```
abramov-oleg: This seems to go against the stated goal of removing extra calls. I suggest not wrapping the logger calls into another function. 

If saving a few lines of code is important, I'd say it is better to remove the `isEnabledFor` condition check entirely, since we do not actually have any expensive args to evaluate in those log messages.

abramov-oleg (changes requested):

abramov-oleg (approved):

PR #610, opened 2024-01-11
Substitution patterns for image and video files sink

### LINKED ISSUE #609: Substitution parameter not supported in Image File Sink and Video File Sink
The documentation for Image File Sink and Video File Sink says:

>` DIR_LOCATION`: a location to write files to; can be a regular path or a path template; supported substitution parameters are `%source_id` and `%src_filename`;

But only JSON Metadata Sink supports this. In the code, I don't see any pattern replacement being done for Image File Sink and Video File Sink.

Y-T-G:
Closes #609

### DIFF
--- a/adapters/gst/gst_plugins/python/video_files_sink.py
+++ b/adapters/gst/gst_plugins/python/video_files_sink.py
@@ -4,7 +4,12 @@ from typing import Dict, Optional, Union
 from savant_rs.primitives import EndOfStream, VideoFrame
 
 from adapters.python.sinks.chunk_writer import ChunkWriter, CompositeChunkWriter
-from adapters.python.sinks.metadata_json import MetadataJsonWriter, Patterns
+from adapters.python.sinks.metadata_json import (
+    MetadataJsonWriter,
+    Patterns,
+    get_tag_location,
+    get_location,
+)
 from gst_plugins.python.savant_rs_video_demux_common import FrameParams, build_caps
 from savant.api.enums import ExternalFrameType
 from savant.api.parser import convert_ts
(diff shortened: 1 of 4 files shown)


tomskikh (changes requested):
Video File Sink should always write frames from different sources to different files. Writing video frame from different sources to the same video file could lead to a broken video or a video with artefacts. I suggest to make `%source_id` required when `DIR_LOCATION` conatains plceholders (i.e. `%src_filename`). If `DIR_LOCATION` doesn't contain any placeholder, append `%source_id` to it.


If the `DIR_LOCATION` has `%src_filename`, the Video Files Sink creates only the directory only for the first filename from each source.

Here's the directories created by the Video Files Sink with vatious `DIR_LOCATION`:

```
$ tree -d results/issue-609/
results/issue-609/
├── no-template
├── only-filename
│   ├── AVG-TownCentre
│   └── deepstream_sample_720p
├── only-source-id
│   ├── test-1
│   └── test-2
└── source-id-and-filename
    ├── test-1
    │   └── AVG-TownCentre
    └── test-2
        └── deepstream_sample_720p

12 directories
```

[issue-609-sample.zip](https://github.com/insight-platform/Savant/files/13916369/issue-609-sample.zip)

tomskikh:
> If the `DIR_LOCATION` has `%src_filename`, the Video Files Sink creates only the directory only for the first filename from each source.

Image Files Sink has the same issue.

Y-T-G:
@tomskikh 

Do you mean something like this:

```python
            src_location = get_tag_location(frame)
            if Patterns.SOURCE_ID in self.location or Patterns.SRC_FILENAME in self.location:
                if Patterns.SOURCE_ID in self.location:
                    base_location = get_location(self.location, frame.source_id, src_location)
                else:
                    self.logger.error(f"{Patterns.SOURCE_ID} is required when {Patterns.SRC_FILENAME} is present.")        
                    return Gst.FlowReturn.ERROR
            else:
                base_location = os.path.join(self.location, frame.source_id)
```

tomskikh:
@Y-T-G No. Check `DIR_LOCATION` at the start of the sink: for image files sink in the method `main`, for video files sink in the script `video_files.sh` or in the method `VideoFilesSink.do_set_property`.

bwsw:
Will be fixed in #867 
Closed.

PR #622, opened 2024-01-18
Switch to Savant-Rs 0.2

### LINKED ISSUE #612: Switch to Savant-Rs 0.2.9
Use embedded ZMQ implementation.
Frame Attributes constructor API slightly changed.

https://github.com/insight-platform/savant-rs/compare/0.1.87...0.2.9

ZMQ: `python/primitives/zmq_*`

tomskikh:


adapters/gst/sinks/video_files.py
```diff
@@ -0,0 +1,412 @@
...
+from savant.gstreamer import GLib, Gst, GstApp
+from savant.gstreamer.codecs import Codec
+from savant.utils.logging import get_logger, init_logging
+from savant.utils.zeromq import ZeroMQMessage, ZeroMQSource
+
+LOGGER_NAME = 'adapters.video_files_sink'
+DEFAULT_CHUNK_SIZE = 10000
+
+
+# Modified version of savant.gstreamer.runner.GstPipelineRunner
```
denisvmedyantsev, line 26: It might be better to single out a common base version and inherit from it?

adapters/gst/sinks/video_files.sh
```diff
@@ -1,41 +1,4 @@
 #!/bin/bash
```
denisvmedyantsev, line 1: Do we still need this script?

libs/gstsavantframemeta/gstsavantframemeta/src/savantrsprobes.cpp
```diff
@@ -8,6 +8,10 @@
 #include "gstnvdsmeta.h"
 #include "nvdsmeta.h"
 
+
+typedef struct Arc_Vec_BorrowedVideoObject {
```
denisvmedyantsev, line 12: What this structure is for? I don't see it being used for.

denisvmedyantsev (approved):

libs/gstsavantframemeta/gstsavantframemeta/src/savantrsprobes.cpp
bwsw: @denisvmedyantsev this is a stub for compilation purposes.

adapters/gst/sinks/video_files.sh
tomskikh: I've kept it so the old docker-compose files would still work.

PR #626, opened 2024-01-24
Support frame color format other than rgba

### LINKED ISSUE #618: Support NV12 when RGBA is not needed
Many pieces do not use RGBA. If only NVinfer and Nvidia-related SDKs are used, NV12 is usually the way to go. This feature disables the use of our drawfunc functionality and OpenCV CUDA. 

The user must activate the function by specifying it mandatory in the YAML configuration file:

```yaml
parameters:
  frame:
    pixel_format: nv12|rgba
```

When `nv12` is chosen, 

- a non-void `draw_func` specification must result in a error;
- a GpuMat requests must result in a meaningful error (install a new mock func with log and raise);
- image **pre**processing must result in a error (install a new mock func with log and raise).

denisvmedyantsev:


bwsw:
We will not merge it because it does not improve the performance.

PR #630, opened 2024-01-30
623: Implemented mock evicted pad

bwsw:


bwsw:
Do not merge, for troubleshooting.

PR #636, opened 2024-02-02
614 make queue accessible from pyfunc

### LINKED ISSUE #614: Make Queue stats accessible from PyFunc
To estimate bursts, we need to have an option to access the stats of all pyfuncs positioned before them. If the queues are not configured, return None.

```python
queues = self.get_queues()
queue = self.get_queue("name")
q.get_params(*param_names)
```

denisvmedyantsev:


savant/base/pyfunc.py
```diff
@@ -70,6 +71,34 @@ def process_buffer(self, buffer: Gst.Buffer):
...
+            if queue_length > 10:
+                self.logger.warning('Queue %s is full!', queue.get_name())
+        ```
+        """
+        if self.gst_element is not None:
+            gst_pipeline = self.gst_element.get_parent()
+            return get_elements(
+                gst_pipeline,
+                factory='queue',
+                before_element=self.gst_element,
```
bwsw: @denisvmedyantsev Why this limitation is applied? Do not you think that it can be important to know not only queues before a particular element, but also after?

The same, if we return queues, we need a concrete type for them.

savant/base/pyfunc.py
```diff
@@ -70,6 +71,34 @@ def process_buffer(self, buffer: Gst.Buffer):
...
+            )
+        return []
+
+    def get_queue(self, name: str) -> Optional[Gst.Element]:
```
bwsw: @denisvmedyantsev, this implementation is probably suboptimal. Does GStreamer not allow getting exact elements by name? If not, do we need to keep a separate reference table for them? I do not like that we iterate over elements just to find the required. Or, at least cache them in order to avoid excessive iterations.

I also do not like the exposing Gst.Element as a resulting type. If we return a queue, then we need a type representing a queue.

bwsw (changes requested):

savant/base/pyfunc.py
denisvmedyantsev: From the issue description "we need to have an option to access the stats of all pyfuncs positioned before them", I thought here should be "queue" instead of "pyfunc" -> "we need to have an option to access the stats of all queues positioned before the pyfunc". No?

"The same, if we return queues, we need a concrete type for them." - I don't get it, please clarify.

savant/base/pyfunc.py
denisvmedyantsev: Yes, we can use `get_by_name`, I'll change the implementation.
The function is expected to be called once, e.g. on_start. Looking for queues while processing each buffer will be a bad idea anyway. 

"I also do not like the exposing Gst.Element as a resulting type. If we return a queue, then we need a type representing a queue." - python doesn't have Gst.Queue, it's not exactly a type... I don't think we have something to replace Gst.Element here.

savant/base/pyfunc.py
denisvmedyantsev: Now pyfunc has methods: 
- get_queues() returns all queues in the pipeline
- get_queue(name) returns queue by name (any queue, up or downstream the pyfunc)
- get_upstream_queue() returns the last (the nearest) queue before the pyfunc

savant/base/pyfunc.py
denisvmedyantsev: Changed get_queue(name) implementation.

Added GstQueue wrapper for queue element to simplify access to main properties.

bwsw (approved):
API looks good to me.

dorgun (approved):

Issue #704, opened 2024-03-20
video file sink adapter have not SKIP_FRAMES_WITHOUT_OBJECTS flag bug 


BloomBabe:
According to Savant [documentation](https://docs.savant-ai.io/v0.2.11/savant_101/10_adapters.html#video-file-sink-adapter) video file sink adapter should have SKIP_FRAMES_WITHOUT_OBJECTS flag , but in source file adapters/gst/sinks/video_files.py I could not find it, but I found it in image file sink adapters/python/sinks/image_files.py, where it has additional check:
```
def _write_video_frame(self, video_frame: VideoFrame, content: bytes) -> bool:
        if self.skip_frames_without_objects and not frame_has_objects(video_frame):
            self.logger.debug(
                'Frame %s from source %s does not have objects. Skipping it.',
                video_frame.source_id,
                video_frame.pts,
            )
            return False
   ....
```
I think the same check should have been placed in adapters/gst/sinks/video_files.py, but for some reason it was not added.

PR #712, opened 2024-03-27
raw-rgba to jpeg because in JP60 hardware jpeg is ok + 704

bwsw:
The raw-rgba format is inefficient from the memory transfer perspective.


bwsw:
Closes #704

denisvmedyantsev (approved):

PR #808, opened 2024-07-04
805 implement messagedump player adapter

### LINKED ISSUE #805: Implement MessageDump player adapter
Deliver message dumps created by the buffer adapter to downstream elements. Simple adapter that accepts a playlist with filenames and plays them sequentially according to the timestamps embedded or without sync.

placccebo:


adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+        self._unpacker = None
+        self._next_message_dump_file()
+
+    def read(self) -> Optional[Tuple[int, str, Message, bytes]]:
+        """Unpack the next message from the file."""
+
+        if not self._unpacker:
+            return None
+        message = None
+        while message is None:
```
bwsw, line 48: Rewrite without loop. Logic is difficult to trace.

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+        """Open the next message dump file from the list."""
+
+        if self._message_dump_file:
+            self._message_dump_file.close()
+
+        file_path = self._list_file.readline()
+        if not file_path:
+            return False
+
+        self._message_dump_file = open(file_path.strip(), 'rb')
```
bwsw: Change to `with`, handle the situation with no segment file with logging (error).

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+        self._unpacker = msgpack.Unpacker(self._message_dump_file)
```
bwsw: Handle error with proper logging and termination.

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+        return True
+
+    def __del__(self):
+        if self._message_dump_file:
+            self._message_dump_file.close()
+        if self._list_file:
+            self._list_file.close()
+
+
+class Egress(BaseThreadWorker):
```
bwsw: If it is from another adapter, try reusing it if it is possible. Not sure how it differs. Maybe it is not straightforward, then ignore this.

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+
+        message = self._reader.read()
+
+        if message is None:
+            return None
+
+        ts, topic, meta, content = message
+
+        if self._sync_output:
+            if self._previous_ts is not None:
```
bwsw: It can work incorrectly (slower than expected) because you need to count the time you spend on reading data and sending data (actual work). Basically, if you sent m1 at 'now' and know you need to send the next one at 'now + X' then you sleep only 'now + X - new_now'. I think you understand the idea.

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,162 @@
...
+def main():
+    init_logging()
+    logger.info(get_starting_message('message dump player adapter'))
+    # To gracefully shutdown the adapter on SIGTERM (raise KeyboardInterrupt)
+    signal.signal(signal.SIGTERM, signal.getsignal(signal.SIGINT))
+    config = Config()
+    reader = MessageDumpReader(config.playlist_path)
+    egress = Egress(reader, config)
+
+    try:
```
bwsw, line 162: Probably you may have it more simple, without extra thread.

bwsw (changes requested):

adapters/python/sources/message_dump_player.py
placccebo: They have the writer in common, and it's possible to refactor the buffer to update the workload method and use it here with a different logic for receiving messages, but it seemed to me not enough to generalize, so I could rename it to Player to avoid ambiguity

adapters/python/sources/message_dump_player.py
placccebo: I'm using the thread for the future in case we need to update the adapter to listen to new dump files

adapters/python/sources/message_dump_player.py
placccebo: I cannot handle the case if some files are empty without the loop. Should I throw an exception instead of skipping?

adapters/python/sources/message_dump_player.py
placccebo: `with` will close the file, and Unpacker needs the file to be opened, I'll wrap it with try/except

adapters/python/sources/message_dump_player.py
bwsw: ok

adapters/python/sources/message_dump_player.py
bwsw: no need for that.

adapters/python/sources/message_dump_player.py
```diff
@@ -0,0 +1,185 @@
...
+        """Read the next message from the dump file."""
+
+        message = self._reader.read()
+
+        if message is None:
+            return None
+
+        ts, topic, meta, content = message
+
+        if self._sync_output:
```
bwsw, line 141: Please, move this sleep code out of read function to a proper scope.

bwsw (changes requested):

bwsw (approved):

Issue #786, opened 2024-06-17
Update demos to use auxiliary video streams


bwsw:
- Video super-resolution demo;
- OpenCV background removal demo;

PR #823, opened 2024-07-29
Using auxiliary stream intead of paddings in a few samples

placccebo:


bwsw (changes requested):
SR crashes, looks like a bug in the code:

image

bwsw (approved):

bwsw:
Closes #786

PR #833, opened 2024-08-09
Updating Client SDK to support batch processing

### LINKED ISSUE #810: Support VideoFrameBatch data in the buffer adapter and ClientSDK
Check if the buffer adapter allows working with batches and implement the support if it is missing.

placccebo:


savant/client/runner/source.py
```diff
@@ -144,6 +145,24 @@ def send(self, source: Frame, send_eos: bool = True) -> SourceResult:
 
         return result
 
+    def send_batch(self, source_id: str, content: Batch) -> SourceResult:
```
tomskikh: Make internal methods `_send_frame` and `_send_batch`. In `send` and `send_iter` call either of them. That's needed to allow those methods (and `__call__`) to accept both `VideoFrame` and `VideoFrameBatch`.

tomskikh (changes requested):
Wrong issue linked? https://github.com/insight-platform/Savant/issues/810 is about buffer adapter, not client SDK.

savant/client/runner/source.py
```diff
@@ -205,6 +199,34 @@ def send_shutdown(self, source_id: str, auth: str) -> SourceResult:
...
+        :return: Result of sending the frame.
+        """
+
+        zmq_topic, message, content, result = self._prepare_video_frame(source)
+        self._send_zmq_message(zmq_topic, message, content)
+        logger.debug('Sent video frame %s/%s.', result.source_id, result.pts)
+
+        return result
+
+    def _send_batch(self, source_id: str, source: Batch) -> SourceResult:
```
tomskikh: No need to pass `source_id` separately, you can get `source_id` from `VideoFrameBatch.frames`.

savant/client/runner/source.py
```diff
@@ -107,37 +109,29 @@ def __del__(self):
...
-        self._send_zmq_message(zmq_topic, message, content)
-        logger.debug('Sent video frame %s/%s.', result.source_id, result.pts)
+        if isinstance(source, tuple) and isinstance(source[0], VideoFrameBatch):
+            batch, source_id = source
+            result = self._send_batch(source_id, batch)
+        else:
+            result = self._send_frame(source)
+
         if send_eos:
             self.send_eos(result.source_id)
```
tomskikh: Batch can contain frames from different sources. E.g. when the module needs to process video from 2 cameras synchronously. Methods `send` and `send_iter` should send EOS for each `source_id`. I suggest adding `source_ids` to `SourceResult` and using it.

tomskikh (changes requested):

savant/client/runner/source.py
placccebo: Batch should be related to a single topic. Otherwise, there's no point in using a batch, as we would need to unpack it to send frames based on their source (don't consider sending the same batch to all topics because it's pretty strange)

savant/client/runner/source.py
```diff
@@ -170,41 +166,69 @@ def send_iter(
             for source_id in source_ids:
                 self.send_eos(source_id)
 
-    def send_eos(self, source_id: str) -> SourceResult:
+    def send_eos(self, zmq_topic: str) -> SourceResult:
```
tomskikh: The argument should be named `source_id`, not `zmq_topic` since its sending end of stream, which is identified by `source_id`.

tomskikh (changes requested):

tomskikh (approved):

PR #834, opened 2024-08-09
Fix batched_push_timeout documentation

diogomafra:
The documentation is showing the wrong default value for the parameter `batched_push_timeout`. 

The [documentation](https://github.com/insight-platform/Savant/blob/864920d79621d901cc23af0a5d1838139d631564/docs/source/advanced_topics/0_batching.rst#L26) indicates the default value is `4000000` (`4` seconds), but the [source code](https://github.com/insight-platform/Savant/blob/864920d79621d901cc23af0a5d1838139d631564/savant/deepstream/pipeline.py#L104) has the value `2000`, which matches what I see in the logs when I don't inform the parameter: 
`Pipeline frame processing parameters: { ... 'batched-push-timeout': 2000 ... }`

I'm assuming the code is correct and the documentation has to be updated.


bwsw:
I will review the PR soon. Thank you for your contribution!

bwsw (approved):

PR #921, opened 2025-01-19
920 KVS API sample

bwsw:


samples/key_value_api/README.md
```diff
@@ -0,0 +1,30 @@
+# Key-Value Storage API Demonstration
+
+A pipeline demonstrating how to work with the embedded Key-Value store.
+
+Tested on platforms:
+
+- Nvidia Turing, Ampere
+- Nvidia Jetson Orin family
```
denisvmedyantsev: maybe better to remove orin from tested, below "The demo implemented only for x86.."

samples/key_value_api/docker/Dockerfile.x86
```diff
@@ -0,0 +1,3 @@
+FROM ghcr.io/insight-platform/savant-deepstream:latest
+
+COPY scripts/ /scripts/
```
denisvmedyantsev: You can simply mount the scripts, which will prevent you from building a separate docker.

samples/key_value_api/module.yml
```diff
@@ -0,0 +1,16 @@
+# module name, required
+name: ${oc.env:MODULE_NAME, 'telemetry-demo'}
+
+# base module parameters
+parameters:
+  batch_size: 1
+  output_frame:
```
denisvmedyantsev: There is no need to specify `output_frame`.

denisvmedyantsev (approved):

PR #953, opened 2025-03-14
Retag savant rs services

bwsw:


bwsw (commented):
Test standalone container re-tagging.

PR #963, opened 2025-03-17
Add GCS support

aeroith:
Also adds session token support for mfa enabled aws tokens.


aeroith:
I've tested it on my local and works for our buckets, but let me know if you need adjustments. Thanks!

bwsw:
@aeroith thank you for your PR. We will review it soon.

bwsw (approved):

PR #990, opened 2025-04-28
EOS on start

bwsw:


adapters/gst/sources/ffmpeg.sh
```diff
@@ -34,6 +34,7 @@ SINK_PROPERTIES=(
     socket="${ZMQ_ENDPOINT}"
     sync="${SYNC_OUTPUT}"
     ts-offset="${SYNC_DELAY}"
+    eos-on-start="${EOS_ON_START}"
```
denisvmedyantsev, line 38: There is no value by default. Have you checked the behavior if we don't provide EOS_ON_START?

adapters/gst/sources/rtsp.sh
```diff
@@ -35,6 +35,7 @@ SINK_PROPERTIES=(
     socket="${ZMQ_ENDPOINT}"
     sync="${SYNC_OUTPUT}"
     ts-offset="${SYNC_DELAY}"
+    eos-on-start="${EOS_ON_START}"
```
denisvmedyantsev, line 39: There is no value by default. Have you checked the behavior if we don't provide EOS_ON_START?

denisvmedyantsev (changes requested):

denisvmedyantsev (approved):

Issue #1014, opened 2025-07-16
YOLOV8Face does not address ROI paddings (top, left)


bwsw:
Fix the converter. 
Fix YOLOV5Face too.

Fix typo: `ration_width` to `ratio_width`.

ryan-goosling:
Fixed in #1017

PR #1017, opened 2025-07-18
Fix/yolov8face roi padding #1014

ryan-goosling:
Fixed the YOLOV8Face converter
Fixed YOLOV5Face as well

Corrected typo: ration_width → ratio_width


bwsw (approved):

savant/converter/yolo_v5face.py
denisvmedyantsev, line 1: My suggestion is to remove this converter. It is not used, and I think there is no point in maintaining yolov5 separately.

denisvmedyantsev: The only sample that used yolov5 was age_gender, but now it uses yolov8. Please, correct title for samples/age_gender_recognition/README.md

savant/converter/yolo_v8face.py
```diff
@@ -76,13 +77,22 @@ def __call__(
         xywh = selected_nms_predictions[:, :4]
         conf = selected_nms_predictions[:, 4:5]
         class_num = np.zeros_like(conf)
-        xywh *= np.tile(np.float32([ration_width, ratio_height]), 2)
+
+        # Scale and shift bbox coordinates
+        xywh *= np.tile(np.float32([ratio_width, ratio_height]), 2)
+        xywh[:, 0] += roi_top  # x center
```
denisvmedyantsev: +roi_top for x center? Did you test with different padding to top and left?

denisvmedyantsev: Please check and correct if needed. 

I suggest we do a bit more in this ticket, let's support maintain_aspect_ratio here as it is done for the standard yolo converter.

denisvmedyantsev (changes requested):

savant/converter/yolo_v5face.py
ryan-goosling: -yolov5face converter removed
-readme fixed

savant/converter/yolo_v8face.py
ryan-goosling: - yes, I tested with different paddings, it looks okay
up_left_pad
bot_right_pad
- a separate [ticket](https://github.com/insight-platform/Savant/issues/1018) has already been created for this task

denisvmedyantsev: Ok. But don't you agree that the code and the comment look inconsistent?
I think the problem is here
`roi_top, roi_left, roi_width, roi_height = roi` should be `roi_left, roi_top, roi_width, roi_height = roi`

ryan-goosling: I relied on how it looks in the annotation `:param roi: [top, left, width, height] of the rectangle`. But I agree, it looks like there was a mistake in the description. I fixed it

savant/converter/yolo_v8face.py
```diff
@@ -51,8 +51,9 @@ def __call__(
         """
         attr_name = model.output.attributes[0].name
 
-        ration_width = roi[2] / model.input.shape[2]
-        ratio_height = roi[3] / model.input.shape[1]
+        roi_left, roi_top, roi_width, roi_height = roi
+        ratio_width = roi_width / model.input.shape[2]
```
denisvmedyantsev, line 55: `model.input.shape[2]` -> `model.input.width`
`model.input.shape[1]` -> `model.input.height`
- a little clearer, but it's up to you.

denisvmedyantsev (approved):

PR #1019, opened 2025-07-24
Symmetric padding support in converters #1018

### LINKED ISSUE #1018: Symmetric padding support in converters
Fix converters to support:

```
model.input
          maintain_aspect_ratio: true
          symmetric_padding: true
```

when `model.input.maintain_aspect_ratio` is used, objects are not distorted, but right-bottom padding is added without `symmetric_padding` or padding where required are added with `symmetric_padding` to make the object in the center of the model input.

e.g.

object 40x60, model 120x120,

Without sp:
```
x: 40 + 80 
y: 60 + 60
```

With sp:
```
x: 40 + 40 + 40
y: 30 + 60 + 30
```

Image

ryan-goosling:
This PR fixes [#1018](https://github.com/insight-platform/Savant/issues/1018) by adding support for `maintain_aspect_ratio`  and `symmetric_padding` in YOLO converters.

Changes:

- Updated yolo.py, yolo_v8face.py, yolo_v4.py to correctly handle padding and aspect ratio
- Ensured bounding boxes are mapped back to the ROI properly


samples/traffic_meter/module.yml
```diff
@@ -70,6 +70,49 @@ pipeline:
...
+                converter:
+                  module: savant.converter.yolo_v4
+                  class_name: TensorToBBoxConverter
+                  kwargs:
+                    confidence_threshold: 0.5
+                    top_k: 300
+                objects:
+                  - class_id: ${parameters.detected_object.id}
+                    label: ${parameters.detected_object.label}
+                    selector:
```
denisvmedyantsev: Please use the same selector and args for each detector. I suggest using MinMaxSizeBBoxSelector instead of the default BBoxSelector, since the converter already performs filtering based on confidence and applies NMS.

savant/converter/yolo.py
```diff
@@ -34,6 +39,50 @@ def __init__(
...
 
+    def _decode_single_head(
+        self, output: np.ndarray, num_classes: Optional[int]
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+        """
+        Returns (bboxes_xywh, confidences, class_ids) from a single output tensor.
+        Handles both [N, 4+C]/[N, 5+C] and transposed [(4+C),N]/[(5+C),N].
+        """
+        # Squeeze batch dim if present
+        if output.ndim == 3 and output.shape[0] == 1:
```
denisvmedyantsev: Can you explain the case when we receive batch in dims? If this is not possible, the check should be removed.

savant/converter/yolo.py
```diff
@@ -34,6 +39,50 @@ def __init__(
...
+            output = output[0]
+
+        # Try to detect transposed layout using num_classes if provided
+        def _is_transposed(arr: np.ndarray) -> bool:
+            if num_classes is None:
+                # Heuristic: if first dim is "plausible" channel count (84/85/etc.)
+                return arr.shape[0] in (84, 85, 92, 93)  # typical for COCO and similar
```
denisvmedyantsev: I believe that we cannot use such heuristics. If there is no way to automatically determine the YOLO type, let's introduce the type as a mandatory argument for the converter.

savant/converter/yolo.py
```diff
@@ -56,92 +104,106 @@ def __call__(
...
         bboxes[:, 0] += roi_left
         bboxes[:, 1] += roi_top
 
+        # --- pack result
         return np.concatenate(
             (
                 class_ids.reshape(-1, 1).astype(np.float32),
-                confidences.reshape(-1, 1),
-                bboxes,
+                confidences.reshape(-1, 1).astype(np.float32),
```
denisvmedyantsev: Are you sure we need these type conversions (`astype`)? You specify the type at the very beginning, and it should not change during the ops.

savant/converter/yolo.py
```diff
@@ -56,92 +104,106 @@ def __call__(
...
+        elif bboxes.shape[0] > self.top_k:
+            idx = np.argpartition(confidences, -self.top_k)[-self.top_k :]
+            bboxes, confidences, class_ids = (
+                bboxes[idx],
+                confidences[idx],
+                class_ids[idx],
             )
+
+        # --- map from model input space -> ROI space
+        roi_left, roi_top, roi_w, roi_h = roi
```
denisvmedyantsev: What is the reason to rename `roi_left, roi_top, roi_width, roi_height = roi` to `roi_left, roi_top, roi_w, roi_h = roi`? `roi_width` is cleared than `roi_w`.

savant/converter/yolo.py
```diff
@@ -56,92 +104,106 @@ def __call__(
...
+
+        if model.input.maintain_aspect_ratio:
+            # reverse letterbox: remove padding, then de-scale
+            scale = min(model.input.width / roi_w, model.input.height / roi_h)
+            if scale <= 0:
```
denisvmedyantsev: Can you explain under what conditions the scale can be negative?

savant/converter/yolo.py
```diff
@@ -56,92 +104,106 @@ def __call__(
...
+                return None
+            # de-scale w,h,xc,yc
+            bboxes /= scale
+
+            if getattr(model.input, 'symmetric_padding', False):
```
denisvmedyantsev: Base model input has default value for symmetric_padding (as well as for maintain_aspect_ratio), there is no reason to use `getattr` here.

savant/converter/yolo.py
```diff
@@ -56,92 +104,106 @@ def __call__(
...
+            bboxes, confidences, class_ids = (
+                bboxes[idx],
+                confidences[idx],
+                class_ids[idx],
             )
+
+        # --- map from model input space -> ROI space
+        roi_left, roi_top, roi_w, roi_h = roi
+
+        if model.input.maintain_aspect_ratio:
```
denisvmedyantsev: Can we scale bboxes once at the end?
```
        if model.input.maintain_aspect_ratio:
            scale = min(
                model.input.width / roi_width,
                model.input.height / roi_height,
            )
            if model.input.symmetric_padding:
                bboxes[:, 0] -= (model.input.width - roi_width * scale) / 2
                bboxes[:, 1] -= (model.input.height - roi_height * scale) / 2
            bboxes /= scale
        else:
            ...
```

savant/converter/yolo_v8face.py
```diff
@@ -78,8 +78,29 @@ def __call__(
...
+            if model.input.symmetric_padding:
+                new_width = roi_width * scale
+                new_height = roi_height * scale
+
+                # Convert to ROI coordinates
+                pad_x = (model.input.width - new_width) / (2 * scale)
+                pad_y = (model.input.height - new_height) / (2 * scale)
+
+                xywh[:, 0] -= pad_x  # xc
+                xywh[:, 1] -= pad_y  # yc
```
denisvmedyantsev: Shouldn't we treat landmarks in a similar way?

denisvmedyantsev (changes requested):

denisvmedyantsev (approved):

PR #1049, opened 2025-08-22
Custom absolute timestamp offset

diogomafra:
Hi, we have a system that requires absolute timestamps, but they should start from a predefined time (actual video time) and not the time the adapter launched.

By default, adapters will use the current time when `USE_ABSOLUTE_TIMESTAMPS` is enabled. Now, I added the parameter `ABSOLUTE_TIMESTAMPS_OFFSET` that allows us to define a custom value.

Is this something you're willing to merge? Please feel free to suggest any changes, add commits, or completely rewrite the PR.


gustavovaliati (commented):
Thanks @diogomafra ! This is a quite useful feature.

bwsw:
@diogomafra Thank you for the contribution, I will review and merge if ok.

bwsw (approved):

Issue #1058, opened 2025-09-01
'model_uid' referenced before assignment


denisvmedyantsev:
DRAFT
```savant.config.module_config.ModuleConfigException: Element nvinfer@detector:v1(name=lpr): local variable 'model_uid' referenced before assignment```

PR #1059, opened 2025-09-08
Fix "'model_uid' referenced before assignment" in case of use before declaration.

denisvmedyantsev:


bwsw (approved):

denisvmedyantsev:
Solves part of problem https://github.com/insight-platform/Savant/issues/1058

PR #1061, opened 2025-09-10
1060 buffer reset

### LINKED ISSUE #1060: Buffer reset feature
Introduce the buffer reset feature. The feature allows setting `BUFFER_RESET_ON_RESTART=True/False` and causes cleanup of the buffer during the restarts of the container. It is beneficial when the system must work with the current data after restarts.

bwsw:
Implemented removal of non-current content when the buffer is restarted.
Closes #1060


samples/buffer_adapter/docker-compose.l4t.yml
```diff
@@ -15,7 +15,7 @@ services:
     entrypoint: /opt/savant/adapters/gst/sources/video_loop.sh
     depends_on:
       module:
-        condition: service_healthy
+       condition: service_healthy
```
denisvmedyantsev: looks strange

denisvmedyantsev (approved):

PR #1072, opened 2025-10-01
[WIP DO NOT MERGE] Use DeepStream 7.1

idantene:
While work on integrating DS8.0 is underway, it would be beneficial to at least upgrade Savant to DS7.1 and allow more modern TensorRT (10.3) in current workflows.
Worth noting that TRT 10.3 (and DS7.1) are the latest support for Orin devices, as per the support matrix [here](https://catalog.ngc.nvidia.com/orgs/nvidia/collections/deepstream_sdk).

I'm not sure what else may need to be updated other than this small image tweak.  
The [release notes](https://docs.nvidia.com/metropolis/deepstream/dev-guide/text/DS_Release_notes.html#ds-7-1-previous-release) seem to be mostly subtle changes, with a few new bindings available and an update to pybind11 to v2.13.0.


Chuong98CC:
@idantene did you manage to make the new docker image working?

bwsw:
@Chuong98CC basically in the current MS (0.5.16) we have a task to rebuild Savant Nvinfer with new TRT without upgrading to 7.1 or 8.0 to support Blackwell:

https://github.com/insight-platform/Savant/milestone/27

We now see it as the most promising approach without changing DS. The reason for that is that 8.x does not support Orin and we do not plan to upgrade until it will. So, instead we just want upgrade TRT and rebuild Nvinfer to support Blackwell.

This PR is welcome but if it is not moving soon, I will discard it.

bwsw:
@Chuong98CC basically in the current MS (0.5.16) we have a task to rebuild Savant Nvinfer with new TRT without upgrading to 7.1 or 8.0 to support Blackwell:

https://github.com/insight-platform/Savant/milestone/27

We now see it as the most promising approach without changing DS. The reason for that is that 8.x does not support Orin and we do not plan to upgrade until it will. So, instead we just want upgrade TRT and rebuild Nvinfer to support Blackwell.

This PR is welcome but if it is not moving soon, I will discard it.

idantene:
> @idantene did you manage to make the new docker image working?

I did not. We were under pressure to get something to work and it was easier to set things up using DeepStream's Service Maker API. It does not decouple the flow as well as Savant, but it was good enough to get the ball going.

Chuong98CC:
> So, instead we just want upgrade TRT and rebuild Nvinfer to support Blackwell.

Yes, we really need the TensorRT 10.6 or above to use model with attention such as RF-DETR. Is there any documents that we can build the Docker with different TensorRT version. We tried to change the TensorRT version with the current docker file and it does not work. THanks

bwsw:
This PR is closed due to inactivity. We plan to release Savant 0.6.0, supporting Blackwell, in a couple of days.

PR #1083, opened 2025-10-17
Replace black with ruff

denisvmedyantsev:
Ruff is a modern replacement for unify+isort+black.

It also works like a linter.

unify doesn't work for 3.10+


bwsw (approved):

denisvmedyantsev:
closes https://github.com/insight-platform/Savant/issues/1084

PR #1118, opened 2025-12-10
Update nvinfer config parser

okanlv:
This PR enables one to overwrite class-level confidence and nms thresholds with the values coming from a [config file](https://docs.savant-ai.io/v0.5.14/reference/api/generated/savant.deepstream.nvinfer.model.NvInferAttributeModel.html#savant.deepstream.nvinfer.model.NvInferAttributeModel.config_file). In this way, one could use a single `module.yml` file with different GPUs, where each GPU has its own confidence and nms thresholds. If `pre-cluster-threshold` or `nms-iou-threshold` values are missing for a class in the config file, then the default values defined in the following lines are used for these classes, as it was before
https://github.com/insight-platform/Savant/blob/2f69b148e07157e8c4273df910e816c37d0f55aa/savant/deepstream/nvinfer/file_config.py#L211-L212

### DIFF
--- a/savant/deepstream/nvinfer/file_config.py
+++ b/savant/deepstream/nvinfer/file_config.py
@@ -292,14 +292,21 @@ class NvInferConfig:
         config['class-attrs-all'] = {'pre-cluster-threshold': 1e10}
         # replace class-attrs parameters with selector kwargs
         for obj in model_config.output.objects:
+            key = f'class-attrs-{obj.class_id}'
             class_attrs = {}
             if not obj.selector or 'kwargs' not in obj.selector:
                 obj.selector = NVINFER_DEFAULT_OBJECT_SELECTOR
             for field in NvInferConfig._CLASS_ATTR_MAP:
+                if key in config:
+                    # use existing config value as default
+                    value = config[key].get(field.property_name)
+                    if value is not None:
+                        class_attrs[field.property_name] = value
+                        continue
                 value = obj.selector.kwargs.get(field.name)
                 if value is not None:
                     class_attrs[field.property_name] = value
             if class_attrs:
-                config[f'class-attrs-{obj.class_id}'] = class_attrs
+                config[key] = class_attrs
 
         return config


denisvmedyantsev:
Hi, @okanlv 
You can set all the parameters for each class via a `selector`

yml config
```
       ...
        output:
          objects:
            - class_id: 0
              label: person
              selector:
                module: savant.selector.detector
                class_name: BBoxSelector
                kwargs:
                  confidence_threshold: 0.2
                  nms_iou_threshold: 0.5
                  min_width: 32
                  min_height: 96
                  max_width: 320
                  max_height: 960
```
will give nvinfer config
```
[class-attrs-0]
pre-cluster-threshold = 0.2
nms-iou-threshold = 0.5
detected-min-w = 32
detected-min-h = 96
detected-max-w = 320
detected-max-h = 960
```

denisvmedyantsev:
@okanlv Perhaps I misunderstood the case you're solving. Do you want the specified values ​​from the initial nvinfer config to be written in the resulting one? Can you clarify?

We try to avoid using the nvinfer configuration directly and prefer to specify all parameters in the Savant configuration whenever possible. Perhaps there's another solution for your initial use case. For example, you can change parameters via py-resolver (`${py:...}`), or pass them via environment variables.

okanlv:
> @okanlv Perhaps I misunderstood the case you're solving. Do you want the specified values ​​from the initial nvinfer config to be written in the resulting one? Can you clarify?
> 
> We try to avoid using the nvinfer configuration directly and prefer to specify all parameters in the Savant configuration whenever possible. Perhaps there's another solution for your initial use case. For example, you can change parameters via py-resolver (`${py:...}`), or pass them via environment variables.



@denisvmedyantsev I want to simplify the solutions for a few different approaches with this change. I will give a few examples below.

- Supporting different confidence values on different GPUs with a single `module.yml` file. In case one wants to use different confidence values on different GPUs for the same model, a separate `module.yml` should be created for each GPU, unless I am missing something. Although all parameters other than confidence values will be the same, we need to duplicate these variables into separate `module.yml` files. This method is a bit error-prone, considering one needs to keep all duplicate parts the same after a change.
- When a model is retrained so that the confidence values are updated on a single GPU, it's required to update `module.yml` file for that GPU in the current case. However, we could keep `module.yml` as is and just update `model/config_file: model_config.txt` file. Handling a new class addition to a model after retraining will also be easier in that case.

In general, it could be possible to represent the following
https://github.com/insight-platform/Savant/blob/9ee1319ae63df32415e14255eb383f036e6ed434/samples/peoplenet_detector/module.yml#L80-L131

with this

```yaml
 # define pipeline's main elements 
 elements: 
   # primary detector element, inference is provided by the nvinfer Deepstream element 
   # model type is detector (other available types are: classifier, custom) 
   - element: nvinfer@detector 
     # Model's name in the pipeline, mandatory 
     name: peoplenet 
     # model definition 
     model: 
       remote: 
         url: s3://savant-data/models/peoplenet/peoplenet_deployable_quantized_onnx_v2.6.3.zip 
         checksum_url: s3://savant-data/models/peoplenet/peoplenet_deployable_quantized_onnx_v2.6.3.md5 
         parameters: 
           endpoint: https://eu-central-1.linodeobjects.com 
       model_file: resnet34_peoplenet.onnx 
       batch_size: ${parameters.batch_size} 
       config_file: model_config.txt
       label_file: labels.txt
```
Please let me know if you have additional questions. I'm answering your questions below as well

> Do you want the specified values ​​from the initial nvinfer config to be written in the resulting one

Yes, exactly this.

> For example, you can change parameters via py-resolver (`${py:...}`), or pass them via environment variables.

I would appreciate it if you could point me to an example related to this. I can check that out.

denisvmedyantsev:
The changes you've proposed will only work with the regular nvinfer converter/selector and will not work with a Python custom converter/selector. Specifying parameters via a selector will work in both cases.

Here are the two ways to configure something in the module configuration that I mentioned earlier.
1) You can use a python code to determine the required value 
config
https://github.com/insight-platform/Savant/blob/9ee1319ae63df32415e14255eb383f036e6ed434/samples/yolov8_seg/module/module.yml#L27

python
https://github.com/insight-platform/Savant/blob/9ee1319ae63df32415e14255eb383f036e6ed434/samples/yolov8_seg/module/converter_opt.py#L6

2) You can use environment variables, same sample, but earlier
compose file
https://github.com/insight-platform/Savant/blob/7e2683dc268554a0026d0098da9c7be8a1cb8880/samples/yolov8_seg/docker-compose.l4t.yml#L42

config
https://github.com/insight-platform/Savant/blob/7e2683dc268554a0026d0098da9c7be8a1cb8880/samples/yolov8_seg/module/module.yml#L30

okanlv:
@denisvmedyantsev Thanks, that's a good idea. That solves most of my problems. I'm not sure how I can use that method to handle the addition of a new class for only a single GPU when I want to support multiple GPUs with a single `module.yml` file. Do you have any suggestions for that?

denisvmedyantsev:
> @denisvmedyantsev Thanks, that's a good idea. That solves most of my problems. I'm not sure how I can use that method to handle the addition of a new class for only a single GPU when I want to support multiple GPUs with a single `module.yml` file. Do you have any suggestions for that?

You can specify all required classes and then manage them using parameters based on your use case/GPU. To disable a specific class by default, you can set an unachievable confidence threshold and set the actual confidence threshold only for the required GPU.

okanlv:
Thanks @denisvmedyantsev , I can close this PR then.

PR #1181, opened 2026-08-13
fix: update rtdetr sample for Savant 0.6 / DeepStream 7.1

### LINKED ISSUE #1112: Fix rtdetr sample  for use in Savant 0.6
The sample uses DeepStream-Yolo and therefore requires a separate container for building, which is tied to a specific DS version.

In fact, we don't need DeepStream-Yolo here. We can directly use the model with our converter.

Additionally, we can replace RTDETR with modern DEIM (https://github.com/Intellindust-AI-Lab/DEIMv2).

IIIllllIlIlllII:
Fixes #1112

- Update builder base image from DeepStream 7.0 to 7.1
- Update CUDA_VER from 12.2 to 12.6
- Update DeepStream-Yolo from 5af9da1 to 2894bab (TensorRT 10 support)

The old DeepStream-Yolo commit used deprecated TensorRT 8 APIs (IPluginV2DynamicExt) that no longer compile against TensorRT 10.x shipped with DeepStream 7.1.

Note: the RT-DETR ONNX model hosted on S3 (savant-data) was exported with the old DeepStream-Yolo output format. It needs to be re-exported using the export_rtdetr_pytorch.py script from DeepStream-Yolo master to match the new parser. Without re-export, detections are incorrect.

### DIFF
--- a/samples/rtdetr/README.md
+++ b/samples/rtdetr/README.md
@@ -4,6 +4,18 @@ The sample shows how RT-DETR model can be used in a Savant module.
 
 The detector model was prepared in the ONNX format using instructions from [DeepStream-Yolo repo](https://github.com/marcoslucianops/DeepStream-Yolo/blob/master/docs/RTDETR_PyTorch.md).
 
+ONNX export tested with the following dependencies:
+
+- DeepStream-Yolo commit 2894bab
+- RT-DETR commit 0d0a4f5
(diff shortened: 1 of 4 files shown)


bwsw:
@IIIllllIlIlllII hello. Thank you for your PR. We will review it soon!

abramov-oleg (approved):

